PR #157, opened 2015-01-11
refs #8879 - Pin fast_gettext for hammer

orrabin:


orrabin:
closing since this is going to be reverted in foreman core (https://github.com/theforeman/foreman/pull/2071)

PR #160, opened 2015-02-12
fixes #8980 - update to gettext 3.x, unpin locale

domcleal:
3.1.13 adds an intermediate .edit.po file alongside each .po, which is meant
to be kept outside of SCM and updated by users, whereupon it's merged back into
the .po on the next rake gettext:find execution.

Previously we overwrote all gettext .po file changes with files directly from
Transifex, but now these are downloaded to .edit.po and gettext merges them
back in.

---

I'll do the same for hammer-cli-foreman if this is accepted, should be a copy/paste job.

This restricts handling po updates to Ruby 1.9, but I don't think that's a big deal.  It means we can deal with newer versions of locale at runtime.

This mirrors the core update (https://github.com/theforeman/foreman/pull/2096) which is a bit stuck on plugin support.

### DIFF
--- a/locale/Makefile
+++ b/locale/Makefile
@@ -10,9 +10,10 @@ DOMAIN = hammer-cli
 VERSION = $(shell ruby -e 'require "rubygems";spec = Gem::Specification::load("../hammer_cli.gemspec");puts spec.version')
 POTFILE = $(DOMAIN).pot
 MOFILE = $(DOMAIN).mo
-POFILES = $(shell find . -name '*.po')
+POFILES = $(shell find . -name '$(DOMAIN).po')
 MOFILES = $(patsubst %.po,%.mo,$(POFILES))
 POXFILES = $(patsubst %.po,%.pox,$(POFILES))
+EDITFILES = $(patsubst %.po,%.edit.po,$(POFILES))
 
 %.mo: %.po
 	mkdir -p $(shell dirname $@)/LC_MESSAGES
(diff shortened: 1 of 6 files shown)


locale/Makefile
```diff
@@ -44,22 +38,20 @@ uniq-po:
...
+# Extract strings and update the .pot, prepare .edit.po files
 extract-strings:
 	bundle exec rake gettext:find
 
-reset-po:
-	# merging po files is unnecessary when using transifex.com
-	git checkout -- ../locale/*/*po
+# Merge .edit.po into .po
+update-po:
+	bundle exec rake gettext:find
```
tstrachota: Why do we have two actions that do the same? (`extract-strings` and `update-po`)

tstrachota:
One question in the code. 
Otherwise it looks good and works nicely. :+1:

locale/Makefile
domcleal: Mostly because I'm out of better ideas :)

The rake task does three things internally - it a) extracts strings from source files into the .pot, b) updates .edit.po from .pot, c) updates .po from .edit.po

So in extract-strings, its job is to extract strings from source files into .edit.po, and then we call tx pull.  Then in update-po it's to merge .edit.po back into .po.

The .edit.po intermediate file is a new feature of gettext 3.  I considered fighting it, but decided to just adopt it.

tstrachota: Got it, thanks. I guess we can live with that:)

domcleal: Yeah, it'd be handy to split them up, but I'm slightly wary of calling the internal gettext tasks by their names in case they change.

PR #162, opened 2015-03-03
fixes #8374 - fixing warning of already initialized constant, BZ1195690

komidore64:


tstrachota:
It workarounds the problem but I think the real cause is having the `HelpBuilder` defined inside `AbstractCommand`. I'd rather move the `HelpBuilder` into a separate file and keep the constant there. What do you think @komidore64 and @mbacovsky ?

komidore64:
actually, i think the real problem is that `lib/hammer_cli/abstract.rb` is getting loaded into the interpreter twice. `lib/hammer_cli.rb` requires it [here](https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli.rb#L14) then later `lib/apipie/command.rb` requires it again [here](https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/apipie/command.rb). I've noticed that hammer_cli has a few circular file dependencies, and this is one of them. Ruby `>1.8` seems to handle the problem and doesn't print a warning, but Ruby `1.8` (which is on EL6 machines) throws this warning.

tstrachota:
I thought `require` would prevent it from being loaded twice but according to http://ruby-doc.org/core-1.8.6/Kernel.html#method-i-require it can't cope with relative paths. I guess we should be using `require 'hammer_cli/...'` format for all requires in `lib/` to make it clean.

komidore64:
either that, or only require files downward in the tree of files. what i mean by that is `lib/hammer_cli/apipie.rb` should only require files into the `lib/hammer_cli/apipie/` directory. At the same time, it seems weird for `lib/hammer_cli/apipie/command.rb` to reach outside its `apipie` directory and require something else. Especially when we know that file is already required.

load order starting from `lib/hammer_cli.rb`:
1. load `lib/hammer_cli/i18n.rb
... 
14. load `lib/hammer_cli/abstract.rb`
...
16. load `lib/hammer_cli/apipie.rb` which loads `lib/hammer_cli/apipie/option_builder.rb`, then `lib/hammer_cli/apipie/command.rb`. `apipie/command.rb` then attempts to go back up the tree and require some stuff again, which isn't necessary. This is where it tries to require `lib/hammer_cli/abstract.rb` again.

komidore64:
@tstrachota can i request this still be merged and maybe create a redmine issue for us to revisit our file loading strategy later?

tstrachota:
@komidore64 ok, let's merge it and solve the loading strategy separately. I created http://projects.theforeman.org/issues/9714 to track it.
Can you please update the commit message? The referenced issue is marked as duplicate.

domcleal:
and remove the reference to BZ.

komidore64:
@tstrachota @domcleal great! thanks!

komidore64:
@tstrachota updated

tstrachota:
:+1: thanks, @komidore64

PR #163, opened 2015-03-11
fixes #7453 - Adds man page to hammer

shlomizadok:


hammer.1
```diff
@@ -0,0 +1,160 @@
+.\" Manpage for hammer cli.
+.\" Contact domcleal@redhat.com to correct errors or typos.
```
domcleal: I do hope you're kidding.

hammer.1
```diff
@@ -0,0 +1,160 @@
...
+.TH man 1 "11 March 2015" "1.4" "hammer man page"
```
domcleal: I don't think this file should be stored in this format, as the date, version number etc can be autogenerated, as can the difficult formatting.

hammer.1
```diff
@@ -0,0 +1,160 @@
...
+.B SUBCOMMAND
+subcommand
+.TP
+.B [ARG]
+subcommand arguments
+
+.SH OPTIONS
+.SS "Subcommands:"
+.TP
+.B architecture
```
domcleal: It shouldn't document commands specific to hammer-cli-foreman?

hammer.1
```diff
@@ -0,0 +1,160 @@
...
+.B -s, --server SERVER
+remote system address
+.TP
+.B -u, --username USERNAME
+username to access the remote system
+.TP
+.B -v, --verbose
+be verbose
+.SH BUGS
+No known bugs.
```
domcleal: That's not true!

domcleal:
The strings describing various arguments should already be internationalised, is it possible to reuse them here so we get locale-specific man pages (and extract any extras)?

domcleal:
We use asciidoc in foreman and foreman-installer, you might want to consider that or something like pod2man or ronn.

https://github.com/theforeman/foreman/tree/develop/man

mbacovsky:
@shlomizadok, thanks for this useful update!
Any preferences for packaging? Would it be possible to keep just the `hammer.1.ronn` file in the repo and build man page and html during packaging? is rubygem-ronn available as build requirement in Koji?

Any plans with the localised man pages?

Also according to ronn conventions it seems that the file should be named `hammer.1.ronn`. Correct?

@domcleal, could you please confirm that the style issues were fixed?

shlomizadok:
@mbacovsky - There is a rubygem-ronn, found it here http://rpmfind.net/linux/rpm2html/search.php?query=rubygem-ronn and here http://koji.fedoraproject.org/koji/buildinfo?buildID=535333 and for el6 https://yum.puppetlabs.com/el/6/dependencies/x86_64/rubygem-ronn-0.7.3-1.el6.noarch.rpm
I'll rename the file to hammer.1.ronn and remove the html and man.
Thanks.

man/hammer.1.ronn
```diff
@@ -0,0 +1,64 @@
...
+**-s, --server SERVER**		Remote system address
+
+**-u, --username USERNAME**	Username to access the remote system
+
+**-v, --verbose**		Be verbose
+
+BUGS
+----
+If one of hammer commands doesn't work as you would expect, you can run `hammer -d` ... to get
+full debug output from the loggers. It should give you an idea what went wrong.
```
domcleal: "to get full debug logging."

Remove the last sentence.

man/hammer.1.ronn
```diff
@@ -0,0 +1,64 @@
...
+
+If you have questions, don't hesitate to contact us on `foreman-users@googlegroups.com` or
+the `Freenode#theforeman` IRC channel.
```
domcleal: Use the same phrasing as the Foreman man page please, see my previous link.

man/hammer.1.ronn
```diff
@@ -0,0 +1,64 @@
+Hammer - the CLI tool (not only) for Foreman
+============================================
+
+DESCRIPTION
+-----------
+
+Hammer is a generic [clamp-based](https://github.com/mdub/clamp) CLI framework.
+Hammer-cli provides just the core functionality. The core is extensible using plugins that contain application-specific commands.
+
+This architecture allows for easy customization according to your application. Nearly any Ruby script can be turned into a Hammer command, so the possibilities are endless. You also can easily add custom commands for your specific use, such as bulk actions or admin tasks.
```
domcleal: Is this relevant?

domcleal:
1. how does this get used in a gem installation?
2. i18n again?
3. any integration from hammer itself, e.g. --man?

man/hammer.1.ronn
shlomizadok: Copied from https://github.com/theforeman/hammer-cli/blob/master/README.md
If you feel this is irrelevant I'll remove it.

domcleal: It's a different target audience I'd have thought.

shlomizadok:
1. rubygems (currently) does not support installing man page, this can happen only when packaging (rpm or deb)
2. How do I i18n this?
3. This will require saving the man page itself under man/hammer.1 so we could run ```man man/hammer.1``` when running ```hammer --man```. Does this sound reasonable?

domcleal:
I don't know off the top of my head, that's really for you to figure out an implementation, but I think you need to consider this in more detail rather than just adding a simple page, especially about how this can be expanded to plugins.

re 3) I don't see why you'd necessarily need to pregenerate it, it could be on the fly?

man/hammer.1.ronn
mbacovsky: @shlomizadok, I checked some other man pages. The description usualy shows very brief info on what the command is good for. I'd assume linking the README in the `SEE ALSO` section would be more "standard" solution.

mbacovsky:
I'm not sure about usability of the `--man` option. The main benefit of man page is that if you are lost you type `man <command>` and you have it. I'm not sure how user would guess that he should look for `--man` option. Only if he used `--help` but then he would get almost same information, so it would be pointless. I agree that on the man page is some extra information, but then I'd prefer to extend the --help content.

Also while I expect to have man pages for the commands that I installed via `yum` or `apt`, I do not expect to have man pages for stuff I develop or install from other kind of repositories. 

Did you check how man page localization is done in other packages?

tstrachota:
I don't see much added value in `--man` too and would go the standard way.

@domcleal pointed out important thing - man pages for plugins. The current implementation only shows first level options and isn't very different from what you get from `hammer -h`. I'm afraid it could be confusing for people to find "Hammer is a generic clamp based framework" in the man page. I'd expect information about foreman commands in the context we use hammer. We need to spend a bit more time with the design. I could imagine plugins placing their bits of man page to a designated place and a script from hammer core building the man page out od them (probably on package install time).

shlomizadok:
Okay,
Gonna start this from scratch.
The path I wish to follow is:
1. Create the asciidoc options and subcommand from ```hammer -h```, including all of its plugins subcommand
2. Use the generated asciidoc to create the man page (and use it during packaging)
3. ditch the ```--man``` option as it is similar to ```-h```
Would love to hear your comments.

mbacovsky:
My preferred design would be:
- man page for core as is now with section listing available plugins
  - `man hammer`
- per plugin man page containing overall information and list of subcommands plus man page per subcommand (similar to git man pages):
  - `man hammer-cli-foreman`
  - `man hammer-host`
  - `man hammer-architecture`
  - `man hammer-cli-katello`
  - `man hammer-content-view`
  - .....
I'd try to avoid dynamic man pages as possible. I'd focus on possibility for localization of man pages. 

Makes sense?

tstrachota:
@mbacovsky the git style... I like that.

shlomizadok:
Yeah, I like this approach.
I did a little script which generates a man page with the options and subcommands for hammer.
You need to have asciidoc installed (```yum install asciidoc```).
run ```ruby generate_man.rb```
and then ```man ./hammer.1``` to view the generated man page.
this is a draft, let me know what you think of it.
One question regarding your comment: we'd need to create a man page per each plugin, on the plugin itself, right?

man/generate_man.rb
```diff
@@ -0,0 +1,47 @@
...
+  available = `hammer --version`.split('*')
+  plugins = "Available Plugins\n"
+  plugins << "-----------------\n"
+  available.each do |plugin|
+    plugins << plugin.strip!
+    plugins << "\n\n"
+  end
+  plugins << "\n"
+  plugins
+end
```
mbacovsky: Dynamically generated available plugins won't work, as during build time there are none. I'd prefer plugins manualy listed

man/generate_man.rb
```diff
@@ -0,0 +1,47 @@
...
+
+`hammer -h > options.asciidoc`
+tempfile=File.open("file.tmp", 'w')
+f=File.new("options.asciidoc")
+tempfile << available_plugins
+f.each do |line|
+  tempfile << trim_colon(line)
+  if line.downcase=~/^usage:/ || line.downcase=~/^subcommands:/ || line.downcase=~/^parameters:/ || line.downcase=~/^options:/
```
mbacovsky: For the same reason as above the subcommands will be probably empty so I'd replace it with some hint to list available subcommands with `hammer -h` and that there are man pages for the top-level subcommands e.g. `man hammer-user`

man/generate_man.rb
mbacovsky: Also hint that there is `man <plugin-name>` with more detail about plugin content would help.

mbacovsky:
@shlomizadok, I like this approach much more. It would be great to check if it will work during build in koji and how the process will look like for the plugins.   

Any idea about the localization?

shlomizadok:
@mbacovsky , trying to revive this pull request.
I have moved from dynamic assigning the plugins into a pre-defined list.
Let me know what you think

ohadlevy:
@mbacovsky anything we can do to move this pr forward? thanks!

man/hammer.1.asciidoc
```diff
@@ -17,7 +17,69 @@ DESCRIPTION
...
 Hammer-cli provides just the core functionality. The core is extensible using plugins that contain application-specific commands.
 
-include::options.asciidoc[]
+Available Plugins
+-----------------
+foreman, foreman_admin, foreman_remote_execution, foreman_discovery
+
+foreman_tasks, foreman_bootdisk, foreman_docker
+
+foreman_salt, foreman_ssh, foreman_deployments
```
tstrachota: I'd mention using `hammer --version` to list enabled plugins here.

tstrachota:
Looks good, @shlomizadok. Static pages are imho enough for the hammer core.

Is there any prototype of a plugin man page already? I also wonder how (or if) we'll handle man page translations.

man/hammer.1.asciidoc
```diff
@@ -0,0 +1,95 @@
...
+
+DESCRIPTION
+-----------
+
+Hammer is a generic [clamp-based](https://github.com/mdub/clamp) CLI framework.
+Hammer-cli provides just the core functionality. The core is extensible using plugins that contain application-specific commands.
+
+Available Plugins
+-----------------
+foreman, foreman_admin, foreman_remote_execution, foreman_discovery
```
mbacovsky: I'd use the plugin names format from `hamer --version` for consistency

mbacovsky:
@shlomizadok, thanks for reviving this and I think this is getting the right direction. 
However, I'm still missing some important answers. As a source of information for a user this has very little value without the commands. I'd like to see more complete solution as part of the PR because hammer plugin authors will need some guidelines to follow.

Questions/comments I'd like to discuss: 

1. how, when and by whom the man page is generated?
1. how the man page is packaged? (this may influence previous point significantly) 
1. localization of the man pages
    - are we going to localize man pages? Foreman doesn't seems so 
    - if yes, how? gettext? directly? what tools?
1. are we going to provide e.g. rake task for installing man page when running from git checkout? For testing and being on par with packaged hammer?
1. we need docs for plugin devs to follow
1. we need tools in core for man page creation in plugins
1. how the plugin man page will look like (output of `man hammer-cli-foreman`)?
1. how the command man page will look like (output of `man hammer-host`)?
1. how we add custom content to man pages besides what is in API help? - host create is a good candidate for fine instructions
1. can we share the custom content with what is in `--help`. - help would benefit from possibility to include hammer specific things that are not part of API docs.
    
These are things I'd like to have clear before this is merged. We will see how much of it is hard to figure out, maybe dev-list or perhaps the RFE tool will show as a better place for the discussion.

tstrachota:
Couple of ideas:

ad 3. - localization
If we're going to localize the man pages, I'd use some templates (probably erb) with gettext. Lot of the strings will be common with what we already have in the code. Final localized man pages would be generated by a rake task (or standalone script) that iterates over available languages and spits out translated man page.

ad 4. - installation from dev mode
This can IMHO be solved later. I don't see anything blocking here.

ad 9. - customized pages
I guess the commands will have separate man page file. Or template if we want to reuse the structure/use translations.

ad 10.
I'm not sure I understand. Can you elaborate, please?

mbacovsky:
ad 9,10: By custom content I mean information that is hammer specific and is not part of the API doc. Let's say we want to add some examples to host create command man page. Question no.9 covers how do we do that. Question no.10 suggest to add the same content also to `hammer host --help` and ideally have it just on one place.

ohadlevy:
@shlomizadok @mbacovsky your thoughts?

mbacovsky:
@ohadlevy: it depends on what we are trying to achieve. If we want to be just standard compliant then going the puppet way seems to be quick and easy solution. However original plan was (or I thought it to be) to provide users some up to date and versioned information that are is available in the API docs (mostly hammer specific stuff that has no reason to be in the API doc, examples, etc). From that perspective saying only the obvious about --help seems pointless.

Both options will have to use very different approach but I can imagine that we will add simple static man page now and replace it with more flexible solution later.

shlomizadok:
I'd like to propose a middle approach:
I'd leave in the man page: Name, Synopsis & Description and add at the end:
"See hammer --help" for help on available hammer subcommands"

What do you think?

regarding some of @mbacovsky questions:
1,2: Man page can be created and installed during RPM install
3: No i18n atm
4. A rake task or a bash script? currently `man/generate_man.sh` generates and installs the man page
5,6: We do need that. :)

shlomizadok:
With latest change it looks like:
screenshot from 2016-06-16 13-45-49

mbacovsky:
@shlomizadok thanks for pushing this forward. For me this is content-wise acceptable compromise.

The only thing I'm missing here is the exact plan how the man page is going to be distributed because I think it can influence the way how it is generated a lot.

If we are going to build and install it during RPM install as I believe you suggest hammer-cli will need to have dependency on asciidoc. We would have to implement that also for debs. I'd like to avoid that.

I checked how it is done in foreman https://github.com/theforeman/foreman/blob/develop/Rakefile.dist#L22-48 and I quite like that approach. Could you please convert the shell script and extend our `build` and `install` tasks. If you update also our `gemspec` to include the man dir the hammer gems will have the compiled man pages seamlessly bundled in and during rpm build we will just copy them over to the right places.

When the packaging is clear I'm good to merge it. :ice_cream:

ohadlevy:
can we agree to move this into a RFC and close the current issue / pr?
thanks!

mbacovsky:
@ohadlevy I think it is good idea for the original content-rich man pages. Do you think it makes sense for this light version? It seems we agreed on the man page content and only missing piece is the distribution, so we are a few steps far from merging it.

I've put together patch solving the packaging and distribution based on what is done in main Foreman.  The patch is here http://pastebin.com/u1jbN5kP. If this is a way to go I can make a PR against Shlomi's branch. What do

shlomizadok:
@mbacovsky - I like your solution. I have applied the patch with the rake task. Thanks!

Rakefile
```diff
@@ -2,12 +2,39 @@ require 'rake/testtask'
 require 'bundler/gem_tasks'
 require 'ci/reporter/rake/minitest'
 
+PREFIX = ENV['PREFIX'] || '/usr/local'
+DATAROOTDIR = ENV['DATAROOTDIR'] || "#{PREFIX}/share"
+MANDIR = ENV['MANDIR'] || "#{DATAROOTDIR}/man"
+
+
```
ohadlevy: nitpick: extra line

tstrachota:
+1 for the rake task

tstrachota:
I tested the rake task and it works well when you have `asciidoc` installed. When it's missing I get:
```
$ rake build
a2x -d manpage -f manpage -D man/ man/hammer.1.asciidoc
rake aborted!
Command failed with status (127): [a2x -d manpage -f manpage -D man/ man/hamm...]
/home/vagrant/hammer/hammer-cli/Rakefile:17:in `block in <top (required)>'
/home/vagrant/.rvm/gems/ruby-2.2.4/bin/ruby_executable_hooks:15:in `eval'
/home/vagrant/.rvm/gems/ruby-2.2.4/bin/ruby_executable_hooks:15:in `<main>'
Tasks: TOP => build => man/hammer.1.gz
(See full trace by running task with --trace)
```
which is not very descriptive.

Would you mind catching the exception and print something like "Make sure you have asciidoc installed"? I guess this is something that users might come across quite frequently.

Othervise :+1: from my side, it looks good.

shlomizadok:
Added a test for a2x, which exits if not found. Let me know if it's a good solution

tstrachota:
@shlomizadok works fine, thanks! :+1: to merge from my side

mbacovsky:
:+1: Thanks @shlomizadok!

PR #164, opened 2015-03-12
Fixes #9721 - ignore spaces in key value formatting

tstrachota:
Updates key-value format to ignore spaces unless they are explicitly
enclosed with quotes. Some examples of input and results of parsing
below:

```
--params "key=[ value1, value2 ]"
{'key' => ['value1', 'value2']}
```
```
--params "key=' value with spaces '"
{'key' => ' value with spaces '}
```


mbacovsky:
:+1: Nice! Thanks for the tests @tstrachota. :peach:

PR #165, opened 2015-03-13
Fixes #9742 - version fails with error

tstrachota:
`hammer --version` could fail with "undefined method 'version'" exception when hammer was configured to use plugin that doesn't define `version`.


mbacovsky:
:+1: Thanks @tstrachota!  :snowman:

PR #166, opened 2015-03-16
fixes #9648 - only initialise text domains that have files in the dir

domcleal:
6786f4cf changed the domain to avoid configuring two identical text domains,
but this reverts that and instead makes the availability check more specific
by checking for the .mo files inside the dir.

This prevents /usr/share/locale being added as a domain when the .mo files
haven't been installed there.

It also only configures the first domain that is available.

---

Tests OK on Debian and EL6 for me.

### DIFF
--- a/lib/hammer_cli/i18n.rb
+++ b/lib/hammer_cli/i18n.rb
@@ -79,7 +79,7 @@ module HammerCLI
       end
 
       def available?
-        File.exist?(locale_dir)
+        Dir[File.join(locale_dir, '**', "#{domain_name}.#{type}")].any?
       end
 
       attr_reader :locale_dir, :domain_name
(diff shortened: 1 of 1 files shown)


lib/hammer_cli/i18n.rb
```diff
@@ -156,6 +152,5 @@ def self.clear
 include FastGettext::Translation
 include HammerCLI::I18n::AllDomains
 
-HammerCLI::I18n.add_domain(HammerCLI::I18n::LocaleDomain.new)
-HammerCLI::I18n.add_domain(HammerCLI::I18n::SystemLocaleDomain.new)
-
+domain = [HammerCLI::I18n::LocaleDomain.new, HammerCLI::I18n::SystemLocaleDomain.new].find { |d| d.available? }
```
domcleal: This shouldn't strictly be needed, but it seems a bit more cautious.

mmoll:
Tested on Debian and FreeBSD, works for me. :+1:

mbacovsky:
:+1: Works fine also on Fedora. Thanks @domcleal!

PR #169, opened 2015-05-20
Fixes #10564 - Added support for deprecated options

mbacovsky:


domcleal:
Note, missing # in the commit + title message.

mbacovsky:
Fixed. Thanks @domcleal

tstrachota:
@mbacovsky could you please update the docs? https://github.com/theforeman/hammer-cli/blob/master/doc/creating_commands.md#option-builders It took me some time to find out how to use the feature.

Not sure if it's a problem but this approach does not work when you define the options separately:
```ruby
option '--old', 'OLD', 'Old option'
option "--new", "NEW", "Replacing option", :deprecated => { "--old" => "Use --new instead" }
```
No warning is displayed when you do `hammer command --old=1`.

mbacovsky:
@tstrachota thanks for review. I updated the docs and added one more straightforward notation format.

In your example the `:deprecated` option has to be set on the option being depricated.

With the new notation is should look like:
```ruby 
option '--old', 'OLD', 'Old option', :deprecated => "Use --new instead"
option "--new", "NEW", "Replacing option"
```
or with the extended syntax:
```ruby 
option '--old', 'OLD', 'Old option', :deprecated => { '--old' => "Use --new instead" }
option "--new", "NEW", "Replacing option"
```

tstrachota:
:+1: it's more intuitive now, thanks @mbacovsky

PR #172, opened 2015-06-22
Fixes #10902 - Add missing options to preparser

mbacovsky:
In the following example the preparser in hammer binary silently fail on unknown options like `--csv` and do not parse the rest of the args ( e.g. ceredentials ) 
```
hammer --csv --username admin --password changeme -d architecture list
```

I've added the missing options to match the main parser.

I've also removed the translating of help texts as it makes no sense here.

### DIFF
--- a/bin/hammer
+++ b/bin/hammer
@@ -9,19 +9,26 @@ HighLine.color_scheme = HighLine::SampleColorScheme.new
 require 'hammer_cli/i18n'
 
 require 'hammer_cli/options/normalizers'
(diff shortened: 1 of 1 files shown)


tstrachota:
@mbacovsky I haven't tested it yet but it LGTM so far. Could you please add a comment above the options mentioning why it's not translated? Just to prevent future confusion.

tstrachota:
works fine :+1:

mbacovsky:
Thanks for review @tstrachota. Comment added.

PR #171, opened 2015-06-16
Refs #8810 - send params of type number as numbers

mbacovsky:

PR #173, opened 2015-07-07
Refs #11033 - set Enum normalizer on enumerated values

mbacovsky:


domcleal:
One too many "0"s in the ticket number.

tstrachota:
Nice improvement, @mbacovsky ! Works well.

I have one idea for improvement: update the `description` method to reflect count of allowed alternatives. "One of ..." sounds strange when there's only one value. What do you think?

mbacovsky:
#171 is also related to normalizer clean up and is touching same code so I added it here as a separate commit referencing issue \#11137

tstrachota:
That was quite confusing shuffle. But I understand it saved us one rebase which comes handy today.

:+1: :monkey_face: thanks, @mbacovsky, merging.

PR #174, opened 2015-07-19
Fixes #8015 - Enable adding defaults options for commands.

alongoldboim:
To add default option to your hammer cli commands you need to define the key value pair(e.g organization_id: 3) under global_defaults in the cli_config.yml(the option will be added if the command permits it in its options).
you can specify a particular resource you want the default option will take effected on by placing the key value pair under <resource>_defauls, this can be done either in the cli_config.yml or in any other plugin config file(e.g foreman.yml).
Need to add a generate defaults command according to plugin and a managing command(add,remove).

### DIFF
--- a/lib/hammer_cli/abstract.rb
+++ b/lib/hammer_cli/abstract.rb
@@ -5,7 +5,6 @@ require 'hammer_cli/clamp'
 require 'hammer_cli/subcommand'
 require 'hammer_cli/options/matcher'
 require 'logging'
-
 module HammerCLI
 
   class AbstractCommand < Clamp::Command
@@ -233,10 +232,12 @@ module HammerCLI
     end
 
     def all_options
-      self.class.recognised_options.inject({}) do |h, opt|
-        h[opt.attribute_name] = send(opt.read_method)
-        h
+      @all_options ||= self.class.recognised_options.inject({}) do |hash, opt|
+        hash[opt.attribute_name] = send(opt.read_method)
+        hash[opt.attribute_name] = add_custom_defaults(opt.attribute_name) if hash[opt.attribute_name].nil?
+        hash
       end
+      @all_options
     end
 
     def options
@@ -245,6 +246,14 @@ module HammerCLI
 
     private
 
+    def add_custom_defaults(attr)
+      if context[:defaults]
+        value = context[:defaults].get_defaults(attr)
+        logger.info("Custom default value #{value} was used for attribute #{attr}") if value
+        value
+      end
+    end
+    
     def self.inherited_output_definition
       od = nil
       if superclass.respond_to? :output_definition
--- a/lib/hammer_cli/defaults.rb
+++ b/lib/hammer_cli/defaults.rb
@@ -1,0 +1,83 @@
+require 'hammer_cli/defaults_commands'
+module HammerCLI
+  DEFAULT_FILE = "#{Dir.home}/.hammer/defaults.yml"
+
+  class Defaults
+    class DefaultsError < StandardError; end
+    class DefaultsPathError < DefaultsError; end
+
+    attr_reader :defaults_settings
+
+    def initialize(settings, file_path = nil)
+
+      @defaults_settings = settings
+      @path = file_path || DEFAULT_FILE
+    end
+
+    def register_provider(provider)
+      providers[provider.plugin_name.to_s] = provider
+    end
+
+    def providers
+      @providers ||= {}
+    end
+
+    def delete_default_from_conf(param)
+      conf_file = YAML.load_file(path)
+      conf_file[:defaults].delete(param)
+      write_to_file conf_file
+      conf_file
+    end
+
+    def add_defaults_to_conf(default_options, provider)
+      create_default_file if defaults_settings.nil?
+      defaults = YAML.load_file(path)
+      defaults[:defaults] ||= {}
+      default_options.each do |key, value|
+        key = key.to_sym
+        defaults[:defaults][key] = value ? {:value => value,} : {:provider => provider}
+      end
+      write_to_file defaults
+      defaults
+    end
+
+    def get_defaults(opt)
+      option = opt
+      option = opt.gsub("option_",'') if opt.include? "option_"
+      unless defaults_settings.nil? || defaults_settings[option.to_sym].nil?
+        if defaults_settings[option.to_sym][:provider]
+          providers[defaults_settings[option.to_sym][:provider]].get_defaults(option.to_sym)
+        else
+          defaults_settings[option.to_sym][:value]
+        end
+      end
+    end
+
+    def write_to_file(defaults)
+      File.open(path,'w') do |h|
+        h.write defaults.to_yaml
+      end
+    end
+
+    protected
+
+    attr_reader :path
+
+    def create_default_file
+      if Dir.exist?(File.dirname(@path))
+        new_file = File.new(path, "w")
+        new_file.write ":defaults:"
+        new_file.close
+      else
+        raise DefaultsPathError.new(_("Couldn't create %s please create the path before defaults are enabled.") % path)
+      end
+    end
+  end
+
+  def self.defaults
+    @defaults ||= Defaults.new(HammerCLI::Settings.settings[:defaults])
+
+  end
+
+  HammerCLI::MainCommand.subcommand "defaults", _("Defaults management"), HammerCLI::DefaultsCommand
+end
--- a/lib/hammer_cli/defaults_commands.rb
+++ b/lib/hammer_cli/defaults_commands.rb
@@ -1,0 +1,163 @@
+require 'hammer_cli'
+require 'yaml'
+module HammerCLI
+  class BaseDefaultsProvider
+    def self.plugin_name
+      self.name.split('::').first.gsub(/^HammerCLI/, '').underscore
+    end
+
+    def self.register_provider
+      HammerCLI.defaults.register_provider(self)
+    end
+
+    def self.support?
+      raise NotImplementedError
+    end
+
+    def self.supported_defaults
+      raise NotImplementedError
+    end
+
+    def self.get_defaults
+      raise NotImplementedError
+    end
+  end
+
+  class DefaultsCommand < HammerCLI::AbstractCommand
+    class ProvidersDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'providers'
+      desc _('List all the providers')
+
+      def execute
+        data = context[:defaults].providers.map do |key, val|
+          {
+            :provider => key.to_s,
+            :defaults => (val.supported_defaults || ['*']).map(&:to_s)
+          }
+        end
+
+        fields = HammerCLI::Output::Dsl.new.build do
+          field :provider, _('Provider')
+          field :defaults, _('Supported defaults'), Fields::List
+        end
+
+        definition = HammerCLI::Output::Definition.new
+        definition.append(fields)
+
+        print_collection(definition, data)
+        HammerCLI::EX_OK
+      end
+
+      def adapter
+        @context[:adapter] || :table
+      end
+    end
+
+    class ListDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'list'
+      desc _('List all the default parameters')
+
+      def execute
+        data = context[:defaults].defaults_settings.map do |key, val|
+          {
+            :parameter => key.to_s,
+            :value     => val[:provider] ? "Provided by: " + val[:provider].to_s.capitalize : val[:value]
+          }
+        end
+
+        fields = HammerCLI::Output::Dsl.new.build do
+          field :parameter, _('Parameter')
+          field :value, _('Value'), Fields::List
+        end
+
+        definition = HammerCLI::Output::Definition.new
+        definition.append(fields)
+
+        print_collection(definition, data)
+        HammerCLI::EX_OK
+      end
+
+      def adapter
+        @context[:adapter] || :table
+      end
+    end
+
+    class DeleteDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'delete'
+
+      desc _('Delete a default param')
+      option "--param-name", "OPTION_NAME", _("The name of the default option"), :required => true
+
+      def execute
+        if context[:defaults].defaults_settings && context[:defaults].defaults_settings[option_param_name.to_sym]
+          context[:defaults].delete_default_from_conf(option_param_name.to_sym)
+          param_deleted(option_param_name)
+        else
+          variable_not_found
+        end
+        HammerCLI::EX_OK
+      end
+    end
+
+    class AddDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'add'
+
+      desc _('Add a default parameter to config')
+      option "--param-name", "OPTION_NAME", _("The name of the default option (e.g. organization_id)."), :required => true
+      option "--param-value", "OPTION_VALUE", _("The value for the default option")
+      option "--plugin-name", "OPTION_PLUGIN_NAME", _("The name of the provider providing the value. For list available providers see `hammer defaults providers`.")
+
+      def execute
+        if option_plugin_name.nil? && option_param_value.nil? || !option_plugin_name.nil? && !option_param_value.nil?
+          bad_input
+          HammerCLI::EX_USAGE
+        else
+          if option_plugin_name
+            namespace = option_plugin_name
+            if !context[:defaults].providers.key?(namespace)
+              plugin_prob_message(namespace)
+              return HammerCLI::EX_USAGE
+            elsif !context[:defaults].providers[namespace].support?(option_param_name)
+              defaults_not_supported_by_provider
+              return HammerCLI::EX_CONFIG
+            end
+          end
+          context[:defaults].add_defaults_to_conf({option_param_name => option_param_value}, namespace)
+          added_default_message(option_param_name.to_s, option_param_value)
+          HammerCLI::EX_OK
+        end
+      rescue Defaults::DefaultsError, SystemCallError => e
+        print_message(e.message)
+        HammerCLI::EX_CONFIG
+      end
+    end
+
+    def added_default_message(key, value)
+      print_message(_("Added %{key_val} default-option with value that will be generated from the server.") % {:key_val => key.to_s}) if value.nil?
+      print_message(_("Added %{key_val} default-option with value %{val_val}.") % {:key_val => key.to_s, :val_val => value.to_s}) unless value.nil?
+    end
+
+    def plugin_prob_message(namespace)
+      print_message(_("Provider #{namespace} was not found. See `hammer defaults providers` for available providers."))
+    end
+
+    def defaults_not_supported_by_provider
+      print_message(_("The param name is not supported by provider. See `hammer defaults providers` for supported params."))
+    end
+
+    def param_deleted(param)
+      print_message(_("%{param} was deleted successfully.") % {:param => param.to_s})
+    end
+
+    def bad_input
+      print_message(_("You must specify value or a provider name, cant specify both."))
+    end
+
+    def variable_not_found
+      print_message(_("Couldn't find the requested param in %s.") % context[:defaults].send(:path))
+    end
+
+    autoload_subcommands
+  end
+end
+
--- a/lib/hammer_cli/settings.rb
+++ b/lib/hammer_cli/settings.rb
@@ -18,6 +18,7 @@ module HammerCLI
         if File.directory? full_path
           # check for cli_config.yml
           load_from_file(File.join(full_path, 'cli_config.yml'))
+          load_from_file(File.join(full_path, 'defaults.yml'))
           # load config for modules
           Dir.glob(File.join(full_path, 'cli.modules.d/*.yml')).sort.each do |f|
             load_from_file(f)
--- a/test/functional/defaults_test.rb
+++ b/test/functional/defaults_test.rb
@@ -1,0 +1,206 @@
+require File.join(File.dirname(__FILE__), '../unit/test_helper')
+describe 'commands' do
+
+  class TestProvider < HammerCLI::BaseDefaultsProvider
+    def self.support?(param)
+      param.to_s == 'organization_id'
+    end
+
+    def self.get_defaults(param)
+      32
+    end
+
+    def self.supported_defaults
+      ['organization_id']
+    end
+  end
+
+  before do
+    settings  = YAML::load(File.open(FILEPATH))
+    @defaults = HammerCLI::Defaults.new(settings[:defaults], FILEPATH)
+    @defaults.stubs(:write_to_file).returns true
+    @defaults.stubs(:providers).returns({
+                                          'foreman' => TestProvider
+                                        })
+
+    @context = {
+      :defaults => @defaults
+    }
+  end
+
+  def expected_message(header, data = [])
+    (header.join("\n") + data.join("\n"))
+  end
+
+  def run_cmd(cmd_class, options)
+    capture_io do
+      cmd_class.run('hammer', options, @context)
+    end
+  end
+
+  describe 'defaults list' do
+
+    it 'it prints all defaults' do
+      header         = ['----------------|----------------------------------------',
+                        'PARAMETER       | VALUE                                  ',
+                        '----------------|----------------------------------------',
+                        ''
+      ]
+      default_values = {
+        :organization_id => {
+          :value => 3,
+        },
+        :location_id     => {
+          :provider => 'HammerCLIForeman::Defaults'
+        }
+      }
+      @defaults.stubs(:defaults_settings).returns(default_values)
+      data     = [
+        'organization_id | 3                                      ',
+        'location_id     | Provided by: Hammercliforeman::defaults',
+        '----------------|----------------------------------------',
+        ''
+      ]
+      out, err = run_cmd(HammerCLI::DefaultsCommand::ListDefaultsCommand, [])
+      assert_equal "", err
+      assert_equal expected_message(header, data), out
+    end
+
+    it 'prints empty defaults' do
+      header = ['----------|------',
+                'PARAMETER | VALUE',
+                '----------|------',
+                ''
+      ]
+      @defaults.stubs(:defaults_settings).returns({})
+
+      out, err = run_cmd(HammerCLI::DefaultsCommand::ListDefaultsCommand, [])
+      assert_equal "", err
+      assert_equal expected_message(header), out
+    end
+  end
+
+  describe 'defaults providers' do
+    header = ['---------|-------------------',
+              'PROVIDER | SUPPORTED DEFAULTS',
+              '---------|-------------------',
+              ''
+    ]
+
+    it 'prints all providers and their supported defaults' do
+      data     = ['foreman  | organization_id   ',
+                  '---------|-------------------',
+                  ''
+      ]
+      out, err = run_cmd(HammerCLI::DefaultsCommand::ProvidersDefaultsCommand, [])
+      assert_equal "", err
+      assert_equal expected_message(header, data), out
+    end
+
+    it 'prints empty providers' do
+      @defaults.stubs(:providers).returns({})
+      out, err = run_cmd(HammerCLI::DefaultsCommand::ProvidersDefaultsCommand, [])
+      assert_equal "", err
+      assert_equal expected_message(header), out
+    end
+
+  end
+
+
+  describe 'defaults add' do
+    it 'adds static default' do
+      options = ['--param-name=param', '--param-value=83']
+
+      @defaults.expects(:add_defaults_to_conf).with({'param' => '83'}, nil).once
+
+      out, err = run_cmd(HammerCLI::DefaultsCommand::AddDefaultsCommand, options)
+      assert_equal "", err
+      assert_equal "Added param default-option with value 83.\n", out
+    end
+
+    it 'adds default from provider' do
+      options = ['--param-name=organization_id', '--plugin-name=foreman']
+
+      @defaults.expects(:add_defaults_to_conf).with({'organization_id' => nil}, 'foreman').once
+
+      out, err = run_cmd(HammerCLI::DefaultsCommand::AddDefaultsCommand, options)
+      assert_equal "", err
+      assert_equal "Added organization_id default-option with value that will be generated from the server.\n", out
+    end
+
+    it 'reports unsupported option' do
+      options = ['--param-name=unsupported', '--plugin-name=foreman']
+
+      out, err = run_cmd(HammerCLI::DefaultsCommand::AddDefaultsCommand, options)
+      assert_equal "", err
+      assert_equal "The param name is not supported by provider. See `hammer defaults providers` for supported params.\n", out
+    end
+
+    it 'reports missing parameter name' do
+      options = ['--param-value=83']
+
+      out, err = capture_io do
+        HammerCLI::DefaultsCommand::AddDefaultsCommand.run('hammer', options, @context)
+      end
+      assert_match "option '--param-name' is required", err
+      assert_equal "", out
+    end
+
+    it 'reports missing parameter value or source' do
+      options = ['--param-name=organization_id']
+
+      out, err = run_cmd(HammerCLI::DefaultsCommand::AddDefaultsCommand, options)
+      assert_equal "", err
+      assert_match "You must specify value or a provider name, cant specify both.", out
+    end
+
+    it 'reports unknown plugin' do
+      options = ['--param-name=organization_id', '--plugin-name=unknown']
+
+      out, err = run_cmd(HammerCLI::DefaultsCommand::AddDefaultsCommand, options)
+      assert_equal "", err
+      assert_equal "Provider unknown was not found. See `hammer defaults providers` for available providers.\n", out
+    end
+
+    it 'reports IO errors' do
+      options = ['--param-name=param', '--param-value=83']
+
+      @defaults.expects(:add_defaults_to_conf).raises(Errno::ENOENT, '/unknown/path')
+
+      out, err = run_cmd(HammerCLI::DefaultsCommand::AddDefaultsCommand, options)
+      assert_equal "", err
+      assert_equal "No such file or directory - /unknown/path\n", out
+    end
+  end
+
+  describe 'defaults delete' do
+    it 'removes the defaults' do
+      default_values = {
+        :organization_id => {
+          :value       => 3,
+          :from_server => false
+        }
+      }
+      @defaults.stubs(:defaults_settings).returns(default_values)
+      @defaults.expects(:delete_default_from_conf).once
+
+      options = ['--param-name=organization_id']
+
+      out, err = run_cmd(HammerCLI::DefaultsCommand::DeleteDefaultsCommand, options)
+      assert_equal "", err
+      assert_equal "organization_id was deleted successfully.\n", out
+    end
+
+    it 'reports when the variable was not found' do
+      @defaults.stubs(:defaults_settings).returns({})
+      @defaults.stubs(:path).returns('/path/to/defaults.yml')
+
+      options  = ['--param-name=organization_id']
+      out, err = run_cmd(HammerCLI::DefaultsCommand::DeleteDefaultsCommand, options)
+      assert_equal "", err
+      assert_equal "Couldn't find the requested param in /path/to/defaults.yml.\n", out
+    end
+  end
+
+end
+
--- a/test/unit/apipie/command_test.rb
+++ b/test/unit/apipie/command_test.rb
@@ -111,6 +111,7 @@ describe HammerCLI::Apipie::Command do
     end
 
     it "should perform a call to api when resource is defined" do
+      ctx[:defaults] = stub(:get_defaults => {})
       cmd_run.must_equal 0
     end
   end
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ tags
 # Locale files
 locale/*/*.edit.po
 locale/*/*.po.time_stamp
+.idea/*
--- a/bin/hammer
+++ b/bin/hammer
@@ -122,4 +122,4 @@ HammerCLI::I18n.domains.each do |domain|
   logger.debug "'#{domain.type}' files for locale domain '#{domain.domain_name}' loaded from '#{File.expand_path(domain.locale_dir)}'"
 end
 
-exit HammerCLI::MainCommand.run || HammerCLI::EX_OK
+exit HammerCLI::MainCommand.run(File.basename($0), ARGV, HammerCLI.context) || HammerCLI::EX_OK
--- a/lib/hammer_cli.rb
+++ b/lib/hammer_cli.rb
@@ -13,9 +13,11 @@ require 'hammer_cli/completer'
 require 'hammer_cli/option_builder'
 require 'hammer_cli/abstract'
 require 'hammer_cli/main'
+require 'hammer_cli/context'
 
 require 'hammer_cli/apipie'
 
 # extend MainCommand
 require 'hammer_cli/shell'
+require 'hammer_cli/defaults'
 
--- a/lib/hammer_cli/context.rb
+++ b/lib/hammer_cli/context.rb
@@ -1,0 +1,13 @@
+require 'hammer_cli/defaults'
+
+module HammerCLI
+
+  def self.context
+    {
+      :defaults => HammerCLI.defaults
+    }
+  end
+
+end
+
+
--- a/test/unit/defaults_test.rb
+++ b/test/unit/defaults_test.rb
@@ -1,0 +1,40 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+describe HammerCLI::Defaults do
+
+  FILEPATH = File.join(File.dirname(__FILE__), '/fixtures/defaults/defaults.yml')
+
+  before(:all) do
+    settings = YAML::load(File.open(FILEPATH))
+
+    @defaults = HammerCLI::Defaults.new(settings[:defaults], FILEPATH)
+    @defaults.stubs(:write_to_file).returns true
+  end
+
+  it "Should add a default param to defaults file, without a provider" do
+    defaults_result = @defaults.add_defaults_to_conf({"organization_id"=> 3}, nil)
+    assert_equal 3, defaults_result[:defaults][:organization_id][:value]
+  end
+
+  it "Should add a default param to defaults file, with provider" do
+    defaults_result = @defaults.add_defaults_to_conf({"location_id"=>nil}, :foreman)
+    assert_equal :foreman, defaults_result[:defaults][:location_id][:provider]
+  end
+
+  it "Should remove default param from defaults file" do
+    defaults_result = @defaults.delete_default_from_conf(:organization_id)
+    assert_nil defaults_result[:defaults][:organization_id]
+  end
+
+  it "should get the default param, without provider" do
+    assert_equal 2, @defaults.get_defaults("location_id")
+  end
+
+  it "should get the default param, with provider" do
+    fake_provider = mock()
+    fake_provider.stubs(:plugin_name).returns(:foreman)
+    fake_provider.expects(:get_defaults).with(:organization_id).returns(3)
+    @defaults.register_provider(fake_provider)
+    assert_equal 3, @defaults.get_defaults("organization_id")
+  end
+
+end
--- a/test/unit/fixtures/defaults/defaults.yml
+++ b/test/unit/fixtures/defaults/defaults.yml
@@ -1,0 +1,6 @@
+---
+:defaults:
+  :organization_id:
+    :provider: 'foreman'
+  :location_id:
+    :value: 2


test/unit/apipie/command_test.rb
```diff
@@ -154,4 +156,31 @@ class CommandC < CommandA
 
   end
 
+  context "defaults" do
+
+    before :each do
+      HammerCLI::Settings.load(:global_defaults => {:organization_id => 3})
+    end
+
```
alongoldboim: These tests needs some refinements but I just wanted to check we are on the right track and didn't want to waist a lot of time.

alongoldboim:
[test]

lib/hammer_cli/settings.rb
```diff
@@ -82,6 +82,14 @@ def self.deep_merge!(h, other_h)
     def self.symbolize_hash(h)
       h = h.inject({}) { |sym_hash,(k,v)| sym_hash.update(k.to_sym => v) }
     end
+    
+    def self.get_defaults_by_key(key, option)
```
mbacovsky: I'd prefer to keep settings untouched. Could this be moved to separate class and use Settings just as a storage?

alongoldboim: Sure ill add  a wrapper for it.

mbacovsky:
Thanks for the update. It looks clean so far. I'd like to see the `create` command and handling of the defaults defined on the server as part of the prove of concept, because it may influence the design significantly.

I don't understand the use-case for global_defaults vs. user_defaults, could you explain, please?

I'd also separate the defaults under some key, not to mix with other config stuff:
```yaml
:defaults:
  :organization_id: 3
  :location_id: 2 
```

Probably not necessary to resolve now, but any idea how to handle name vs. id resource definition? What if user wants to `defaults create --name organization --value 'My Org'` is it stored as `organization_id` or `organization`, do we want detect conflict if one wants to store both?

alongoldboim:
OK as for the create commands I still need to implement them I just wanted to make sure I'm going in the right direction first, the create shouldn't problem because i can define it on cli_config under global section or if a resource is given then under the resource.

As for generate defaults for plugins, i thought of making a defaults class (will also be the settings wrapper) and on it to have a method generate_defaults(options={}) that will use the create_defaults command to place the options hash to the config file, all the plugins will be able to override and supply their own hash. 

The global vs resource (user is just an example of a resource), the purpose was to give the users an option to define a default on every request(if the request has the option available) as opposed to placing the default on just one resource.

The key separation that i thought of is the <resource>_defaults so u'd be able to pretty much define it in any of the config files and all the keys will have the _defaults in the end.

About your last point, you think we should deal with that? it wont really be a problem unless the two options have values that don't match. When ever a default value is being used its printed to the screen, so in case of collisions the user will know right away.

meanwhile ill add the defaults class and the create option. tell me if you agree on the design so far.

config/cli_config_example.yml
```diff
@@ -0,0 +1,14 @@
...
+  :history_file: ~/.hammer/history
+:watch_plain: false
+:reload_cache: false
+:log_dir: ~/.hammer/log
+:log_level: error
+:global_defaults:
+  :organization_id: 293
+  :organization_name: Redhat
+:user_defaults:
+  :organization_id: 2
```
mbacovsky: Please nest the options one level deeper so it doesn't mess with other config options. Then the fragile `_defaults` suffix is not needed and it can make human editing and even processing more netural.
```yaml
:defaults:
  :_global:
    :organization_id: 293
    :organization_name: Redhat
  :user:
    :organization_id: 2
```

mbacovsky:
I'm not sure about the 'global' vs 'resource related' distinction. Do we have any real use- case for that? 
What if user wants to use the default org_id in more Katello resources, does she end up with 10 same org_id occurrences in the config? 
Would it be possible to have it just global and wait with adding resource blacklist/whitelist after having some users feedback?

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,133 @@
...
+      value = HammerCLI::Settings.settings[key][option] unless HammerCLI::Settings.settings[key].nil?
+      value = HammerCLI::Settings.settings[:global_defaults][option] if value.nil? && !HammerCLI::Settings.settings[:global_defaults].nil?
+      value
+    end
+
+    private
+    def self.find_valid_conf_file
+      path = nil
+      HammerCLI::Settings.path_history.each do |p|
+        path = p if p.include?("cli_config.yml")
```
mbacovsky: Good idea to use the `path_history`. It lead me to interesting idea:
I use my hammer to manage multiple foreman/sat instances. It would be nice to have different defaults for each of them as the org_id will probably differ. I usually use `hammer -c ~/.hammer/dev.yml commands...` where the dev.yml sets the right url and credentials. 
Not sure how to detect the right file in that case though. Or we could put the defaults to separate file and make the path part of the configuration:
```yaml
:defaults_path: '~/.hammer/dev_defaults.yml' 
```
Then you could create the file for the user if it doesn't exist.
What do you think?

mbacovsky:
Besides the comments above I'd recommend to keep the verbosity as low as possible not to complicate automation. Try to use logger with verbosity of your choice where possible instead of `print_message`. Also make sure all the strings visible to user are wrapped in `_()`
Thanks for the updates, it looks it is on right track :beer:

config/cli_config_example.yml
alongoldboim: ok, can do that.

lib/hammer_cli/defaults.rb
alongoldboim: We can also do that, but the thing is that we said that its better the same file.
I think that for now we can leave it like that, if we would like to change it afterwards, its a small fix. 
How common it is for a regular user to manage multiple instances?

alongoldboim:
OK i agree with you, better putting it in the logger, and ill wrap the strings.
Thanks for the over view, ill make an update soon.

mbacovsky:
I looked again at  the generate defaults command. Would you please explain the purpose and usage of this command? Why `defaults add` couldn't do the job?

alongoldboim:
The defaults add command is for a user to add a defaults setting on his own, the generate is a method that every plugin can override and then it can generate defaults for you(e.g in foreman the generate method will already place org_id and loc_id taken from the web UI) , in every plugin we got for hammer it will be implemented differently.

alongoldboim:
@mbacovsky Made some upgrades, take a look and comment please.
Thanks!

tstrachota:
@alongoldboim @mbacovsky I think we should store the values in a dedicated yaml file otherwise users loose structure and comments from their config when new value is added which is quite annoying. It completely ruined my config file when I was testing the PR.

I feel the saved options should have three states: `no value is set` = the option is not saved at all, `user defined value is set`, `use default from the server`. I understand that making an extra api call for getting the defaults with each command that uses options set to the latter is not very handy. The current solution with a command for loading the defaults should be enough (at least for the start).

I propose one improvement, similar to what @mbacovsky said: Could we merge the behaviour of `hammer defaults add` and `hammer defaults generate`? For example:
```bash
hammer defaults add --param-name xxx --param-value yyy # sets user defined value
hammer defaults add --param-name xxx --default # sets server default value
```
This way users will have more control over what parameters are loaded from the server instead of loading all that is available.

alongoldboim:
@tstrachota @mbacovsky , We could implement the defaults add defaults like you suggested no problem.
Regarding the dedicated file we could do it, we'll need to set the file name in the settings config so it will be loaded to the settings?

mbacovsky:
@alongoldboim, I'd prefer for the defaults to have configurable name with some default value. That would allow me having different sets of defaults for different Foreman/Sat instances.

alongoldboim:
We could do  that but don't you think it would be better to keep it simple and easy and get people responses and then make improvements? 
@ohadlevy For this phase you think it would be better to split it into files or keep it in the cli_config file?

mbacovsky:
@alongoldboim I was talking about the defaults dedicated file name you asked for, my bad

tstrachota:
@alongoldboim I tested both PRs and found some issues:
**1)** The defaults file should be imho created by default for users when it doesn't exist. It took me quite some time to find where to place it.
**2)** The defaults file isn't recognized when it's empty.
**3)** `--plugin-name` should be refused when `--param-val` is provided because it's not used at all in such case.
 ```
hammer defaults add --param-name domain_id  --param-val 111  --plugin-name foreman
Added -domain_id default-option with value -111
```
**4)**  I don't think it's a good idea to mix the saved values with the "from_server" setting. One can easily break it:
```
hammer defaults add --param-name organization_id  --param-val from_server_abc
hammer domain list
Error: wrong constant name abc
```
Have you considered two level structure of the defaults file? Something like:
```yaml
:defaults:
  :organization_id:
    :value: 32
    :from_server: false
  :location_id:
    :from_server: true
```
**5)** Loading organizations from server does not work for me:
```
hammer defaults add --param-name organization_id  --plugin-name foreman
hammer domain list
```
then fails with:
```
NoMethodError (undefined method `[]' for nil:NilClass):
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/defaults.rb:14:in `get_defaults'
    /root/hammer/hammer-cli/lib/hammer_cli/defaults.rb:100:in `get_defaults'
    /root/hammer/hammer-cli/lib/hammer_cli/apipie/options.rb:41:in `block in add_default_options'
    /root/hammer/hammer-cli/lib/hammer_cli/apipie/options.rb:39:in `each'
    /root/hammer/hammer-cli/lib/hammer_cli/apipie/options.rb:39:in `add_default_options'
    /root/hammer/hammer-cli/lib/hammer_cli/apipie/options.rb:5:in `method_options'
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/commands.rb:240:in `request_params'
    /root/hammer/hammer-cli/lib/hammer_cli/apipie/command.rb:43:in `send_request'
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/commands.rb:185:in `send_request'
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/commands.rb:275:in `send_request'
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/commands.rb:310:in `retrieve_all'
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/commands.rb:294:in `execute'
    /root/clamp/lib/clamp/command.rb:68:in `run'
    /root/hammer/hammer-cli/lib/hammer_cli/abstract.rb:23:in `run'
    /root/clamp/lib/clamp/subcommand/execution.rb:11:in `execute'
    /root/clamp/lib/clamp/command.rb:68:in `run'
    /root/hammer/hammer-cli/lib/hammer_cli/abstract.rb:23:in `run'
    /root/clamp/lib/clamp/subcommand/execution.rb:11:in `execute'
    /root/clamp/lib/clamp/command.rb:68:in `run'
    /root/hammer/hammer-cli/lib/hammer_cli/abstract.rb:23:in `run'
    /root/clamp/lib/clamp/command.rb:133:in `run'
    ./bin/hammer:115:in `<main>'
```
**6)** Have you considered letting plugins register a defaults provider instead of trying to find the class? Something like:
```ruby
HammerCLIForeman::Defaults.register_provider(HammerCLIForeman::Defaults.new)
```

lib/hammer_cli/apipie/options.rb
```diff
@@ -34,5 +34,18 @@ def get_option_value(opt_name)
...
     end
 
+    def add_default_options(options, explicit_options)
+      defaults = {}
+      options.each do |option|
+        option = option.name.to_s
+        value = HammerCLI::Defaults.get_defaults(option) unless explicit_options["option_"+option]
+        if value
+          defaults["option_" + option] = value
+          logger.info(_("You are using the following default parameter:  #{option} = #{value}"))
```
tstrachota: Either use % and put arguments outside the gettext function or leave the string untranslated. I'm fine with the latter. We usually don't translate log messages.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,120 @@
...
+      end
+    end
+
+    class AddDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'add'
+
+      desc 'Add a default parameter to config'
+      option "--param-name", "option-name", _("The name of the default option"), :required => true
+      option "--param-val", "option-value", _("The value for the default option")
+      option "--plugin-name", "option-name", _("The plugin name defaults will be generated for")
```
tstrachota: For all options above `"option-name"` should be capital. Also please use underscores instead of dashes.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,120 @@
...
+      def execute
+        list_all_defaults_message
+        HammerCLI::EX_OK
+      end
+    end
+
+    class AddDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'add'
+
+      desc 'Add a default parameter to config'
```
tstrachota: Translation

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,120 @@
+require 'hammer_cli'
+require 'yaml'
+module HammerCLI
+
+  class DefaultsCommand < HammerCLI::AbstractCommand
+
+    class ListDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'list'
+
+      desc 'List all the defaults parameters'
```
tstrachota: Translation

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,120 @@
...
+          file_not_found_message
+          HammerCLI::EX_CONFIG
+        end
+        HammerCLI::EX_OK
+      end
+    end
+
+    def added_default_message(options = {})
+      options.each do |key, val|
+        print_message("Added -" + key.to_s + " default-option with value -"+ val.to_s)
```
tstrachota: Translate the string please. What is purpose od the dashes in front of values?

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,120 @@
...
+      end
+    end
+
+    def plugin_prob_message(options = {})
```
tstrachota: Unused options could be removed. BTW did you place the message methods outside the command classes for any reason?

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,120 @@
...
+      print_message(_("You must specify a value or a plugin name."))
+    end
+
+    def file_not_found_message
+      print_message(_("Couldn't find any cli_config file in the recoginized paths"))
+    end
+
+    def list_all_defaults_message
+      HammerCLI::Settings.settings.each do |key,val|
+        if key.to_s.include?("defaults")
```
tstrachota: Will there be more keys for storing defaults in future? Why not just `HammerCLI::Settings.settings["defaults"]` ?

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,120 @@
...
+          value
+        end
+      end
+    end
+
+    private
+    def self.find_valid_conf_file
+      path = nil
+      HammerCLI::Settings.path_history.each do |p|
+        path = p if p.include?("defaults.yml")
```
tstrachota: Empty files are not recognized due to https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/settings.rb#L39

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,120 @@
...
+          value
+        else
+          value = HammerCLI::Settings.settings[:defaults][option.to_sym]
+          value
+        end
+      end
+    end
+
+    private
+    def self.find_valid_conf_file
```
tstrachota: Could be simplified to `HammerCLI::Settings.path_history.find{ |path| path.include?("defaults.yml") }`

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,139 @@
...
+      end
+    end
+
+    class AddDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'add'
+
+      desc _('Add a default parameter to config')
+      option "--param-name", "Option_name", _("The name of the default option"), :required => true
+      option "--param-val", "Option_value", _("The value for the default option")
+      option "--plugin-name", "Option_name", _("The plugin name defaults will be generated for")
```
tstrachota: All capital, please "Option_name" -> "OPTION_NAME"

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,139 @@
...
+          if option_plugin_name.nil? && option_param_val.nil? || !option_plugin_name.nil? && !option_param_val.nil?
+            bad_input
+          else
+            namespace = ("HammerCLI" + option_plugin_name.split('_').collect!{ |w| w.downcase.capitalize }.join + "::Defaults") if option_plugin_name
+            if option_plugin_name
+              raise NameError unless HammerCLI::Defaults.providers.any? {|p| p.to_s.include? namespace}
+              raise NotImplementedError unless HammerCLI::Defaults.providers[0].support? option_param_name
+            end
+            Defaults.add_defaults_to_conf({option_param_name => option_param_val},option_plugin_name ? HammerCLI::Defaults.providers.find_index(namespace.constantize) : "")
+            added_default_message(option_param_name.to_s, option_param_val ? option_param_val.to_s : "that will be generated from the server")
```
tstrachota: This results in:
```
hammer defaults add --param-name organization_id --plugin-name foreman
Added key_val default-option with value that will be generated from the server.
```
I can see several problems here:
1. "that will be ..." is not translated
2. it would be better to use two independent strings here as some languages can have different word order
3. "key_val" was not substituted. It shoud be part of the string. `"Added %{key_val} ..."`

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,139 @@
...
+module HammerCLI
+
+  class DefaultsCommand < HammerCLI::AbstractCommand
+
+    class ListDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'list'
+
+      desc _('List all the defaults parameters')
+      def execute
+        list_all_defaults_message
```
tstrachota: It only prints plain yaml to output. Would you mind to add some more formatting? I imagine the output as:
```
hammer defaults list
organization_id:  (provided by foreman)
location_id:      1
login:            admin
```
or something similar. Instead of current:
```
hammer defaults list
organization_id:
---
:from_server: true
:provider: 0
```
Btw I'm not sure why the provider is `0` here.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,139 @@
...
+    end
+
+    #this method will be overriden by plugins who wish to have a defaults params.
+    def self.get_defaults(option)
+      unless HammerCLI::Settings.settings[:defaults].nil? || HammerCLI::Settings.settings[:defaults][option.to_sym].nil?
+        if HammerCLI::Settings.settings[:defaults][option.to_sym][:from_server]
+          value =  HammerCLI::Defaults.providers[HammerCLI::Settings.settings[:defaults][option.to_sym][:provider]].get_defaults(option.to_sym)
+          value
+        else
+          value = HammerCLI::Settings.settings[:defaults][option.to_sym][:value] if defaults_file_exists?
```
tstrachota: Looking at the body of `defaults_file_exists?`, ain't it too late to check whether `HammerCLI::Settings.settings[:defaults].nil?` here? You've done it on the line `109`

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,139 @@
...
+
+      def execute
+        begin
+          if option_plugin_name.nil? && option_param_val.nil? || !option_plugin_name.nil? && !option_param_val.nil?
+            bad_input
+          else
+            namespace = ("HammerCLI" + option_plugin_name.split('_').collect!{ |w| w.downcase.capitalize }.join + "::Defaults") if option_plugin_name
+            if option_plugin_name
+              raise NameError unless HammerCLI::Defaults.providers.any? {|p| p.to_s.include? namespace}
+              raise NotImplementedError unless HammerCLI::Defaults.providers[0].support? option_param_name
```
tstrachota: You're checking only the first registered provider here.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,139 @@
...
+      option "--plugin-name", "Option_name", _("The plugin name defaults will be generated for")
+
+      def execute
+        begin
+          if option_plugin_name.nil? && option_param_val.nil? || !option_plugin_name.nil? && !option_param_val.nil?
+            bad_input
+          else
+            namespace = ("HammerCLI" + option_plugin_name.split('_').collect!{ |w| w.downcase.capitalize }.join + "::Defaults") if option_plugin_name
+            if option_plugin_name
+              raise NameError unless HammerCLI::Defaults.providers.any? {|p| p.to_s.include? namespace}
```
tstrachota: Suggestion: It would be simpler and more flexible to let the providers define a method that returns their name and save it into a hash during registration. You don't need to do the namespace harakiri then and only check `@@providers.key?(provider_name)`

tstrachota:
When I delete the file and do `hammer default list` I get:
```
NoMethodError (undefined method `each' for nil:NilClass):
    /root/hammer/hammer-cli/lib/hammer_cli/defaults.rb:72:in `list_all_defaults_message'
    /root/hammer/hammer-cli/lib/hammer_cli/defaults.rb:12:in `execute'
    /root/clamp/lib/clamp/command.rb:68:in `run'
    /root/hammer/hammer-cli/lib/hammer_cli/abstract.rb:23:in `run'
    /root/clamp/lib/clamp/subcommand/execution.rb:11:in `execute'
    /root/clamp/lib/clamp/command.rb:68:in `run'
    /root/hammer/hammer-cli/lib/hammer_cli/abstract.rb:23:in `run'
    /root/clamp/lib/clamp/subcommand/execution.rb:11:in `execute'
    /root/clamp/lib/clamp/command.rb:68:in `run'
    /root/hammer/hammer-cli/lib/hammer_cli/abstract.rb:23:in `run'
    /root/clamp/lib/clamp/command.rb:133:in `run'
    ./bin/hammer:121:in `<main>' 
```

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,139 @@
...
+              raise NotImplementedError unless HammerCLI::Defaults.providers[0].support? option_param_name
+            end
+            Defaults.add_defaults_to_conf({option_param_name => option_param_val},option_plugin_name ? HammerCLI::Defaults.providers.find_index(namespace.constantize) : "")
```
tstrachota: I understand the `0` in the config file now. Storing the index may break when you add a plugin later and the alphabetical order changes. It would be better to store the name directly.

tstrachota:
When you set a default and the user doesn't have it set on the server, any command results in:
```
NoMethodError (undefined method `[]' for nil:NilClass):
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/defaults.rb:23:in `get_defaults'
    /root/hammer/hammer-cli/lib/hammer_cli/defaults.rb:111:in `get_defaults'
    /root/hammer/hammer-cli/lib/hammer_cli/apipie/options.rb:40:in `block in add_default_options'
    /root/hammer/hammer-cli/lib/hammer_cli/apipie/options.rb:38:in `each'
    /root/hammer/hammer-cli/lib/hammer_cli/apipie/options.rb:38:in `add_default_options'
    /root/hammer/hammer-cli/lib/hammer_cli/apipie/options.rb:5:in `method_options'
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/commands.rb:240:in `request_params'
    /root/hammer/hammer-cli/lib/hammer_cli/apipie/command.rb:43:in `send_request'
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/commands.rb:185:in `send_request'
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/commands.rb:275:in `send_request'
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/commands.rb:310:in `retrieve_all'
    /root/hammer/hammer-cli-foreman/lib/hammer_cli_foreman/commands.rb:294:in `execute'
    /root/clamp/lib/clamp/command.rb:68:in `run'
    /root/hammer/hammer-cli/lib/hammer_cli/abstract.rb:23:in `run'
    /root/clamp/lib/clamp/subcommand/execution.rb:11:in `execute'
    /root/clamp/lib/clamp/command.rb:68:in `run'
    /root/hammer/hammer-cli/lib/hammer_cli/abstract.rb:23:in `run'
    /root/clamp/lib/clamp/subcommand/execution.rb:11:in `execute'
    /root/clamp/lib/clamp/command.rb:68:in `run'
    /root/hammer/hammer-cli/lib/hammer_cli/abstract.rb:23:in `run'
    /root/clamp/lib/clamp/command.rb:133:in `run'
    ./bin/hammer:121:in `<main>' 
```

tstrachota:
@alongoldboim I'm sorry but the PR still contains serious issues. I can give you a hand and send a PR with updates against your branch next week.

Please use separate commits if you're going to address the issues I mentioned so that we don't loose history of the comments with rebases. We'll squash them just before merge. Thanks!

tstrachota:
One more issue found: When I set `organization_id` to some value manually and then call `hammer user list` the default doesn't seem to be reflected even though the command takes option `--organization_id`.

tstrachota:
:+1: for improvements in the parameter handling. It seems to be working good now and complains about missing params correctly.

alongoldboim:
@tstrachota No need to be sorry :), ill make the needed changes. thanks for the review.
We are getting closer :+1:

lib/hammer_cli/defaults.rb
alongoldboim: i kept it in an array(no real reason for hash i think), but i removed the namespace.constanize.

tstrachota:
@alongoldboim looks good in terms of functionality, reviewing the code now...

test/unit/apipie/command_test.rb
```diff
@@ -154,4 +154,27 @@ class CommandC < CommandA
...
+  #     cmd_opt = OptionCommand.new("", ctx)
+  #     params = cmd_opt.class.resource.action(:create).params
+  #     cmd_opt.method_options({"option_organization" => "1"}).must_equal({"architecture"=>{"name"=>"test_name"}, "organization" =>"1"})
+  #   end
+  #
+  #   def teardown
+  #     HammerCLI::Settings.clear
+  #   end
+  #
+  #   end
```
tstrachota: Hehe, I see that the tests are green :smile:

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
+require 'hammer_cli'
+require 'yaml'
+module HammerCLI
+
+  class DefaultsCommand < HammerCLI::AbstractCommand
+
+    class ListDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'list'
+
+      desc _('List all the defaults parameters')
```
tstrachota: 'defaults' typo?. Should probably be either 'List all default parameters' or 'List all the defaults'.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+      option "--param-val", "OPTION_VALUE", _("The value for the default option")
+      option "--plugin-name", "OPTION_PLUGIN_NAME", _("The plugin name defaults will be generated for")
+
+      def execute
+        begin
+          if option_plugin_name.nil? && option_param_val.nil? || !option_plugin_name.nil? && !option_param_val.nil?
+            bad_input
+          else
+            if option_plugin_name
+              namespace = ("HammerCLI" + option_plugin_name.split('_').collect!{ |w| w.downcase.capitalize }.join + "::Defaults").to_sym
```
tstrachota: You can use the utility method `camelize` for the namespace:
https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/utils.rb#L18

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+              raise NameError unless HammerCLI::Defaults.providers.keys.include?(namespace)
+              raise NotImplementedError unless HammerCLI::Defaults.providers[namespace].support? option_param_name
+            end
+            Defaults.add_defaults_to_conf({option_param_name => option_param_val}, namespace)
+            added_default_message(option_param_name.to_s, option_param_val)
+          end
+          rescue NameError => e
```
tstrachota: Wrong indentation here.

test/unit/apipie/command_test.rb
alongoldboim: its about time no? :dancer:

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+        print_message(_("No defaults file was found"))
+      end
+    end
+
+    autoload_subcommands
+  end
+
+
+  class Defaults
+    def self.register_provider(provider)
```
tstrachota: Use separate variables for the provider name and the provider, please. It will increase readability of the code.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+      Defaults.providers[provider.keys[0]] = provider.values[0]
+    end
+
+    def self.providers
+      @@providers ||= {}
+    end
+
+    def self.add_defaults_to_conf(options, provider)
+      create_default_file unless !HammerCLI::Settings.settings[:defaults].nil?
+      path = "#{Dir.home}/.hammer/cli.modules.d/defaults.yml"
```
tstrachota: The defaults file path and name are used at multiple places. It would be good to define them only once as constants.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+      new_file = YAML.load_file(path)
+      options.each do |key, value|
+        key = key.to_sym
+        value.to_s.to_i if value.is_a? Integer
```
tstrachota: This line has actually no effect.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+      print_message(_("The param name is not supported by plugin"))
+    end
+
+
+    def bad_input
+      print_message(_("You must specify value or a plugin name, cant specify both."))
+    end
+
+    def self.file_cant_be_created
+      print_message(_("Couldn't add file to ~/.hammer/cli.modules.d please create the path before defaults will be enabled."))
```
tstrachota: Please use %s substitution for the config path.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    def self.create_default_file
+      path = "#{Dir.home}/.hammer/cli.modules.d/"
+      if Dir.exist?(path)
+        new_file = File.new(path + "/defaults.yml", "w")
+        new_file.write ":defaults:"
+        new_file.close
+      else
+        DefaultsCommand.file_cant_be_created
```
tstrachota: What's the reason for keeping the messages in a different class?

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+        h.write new_file.to_yaml
+      end
+    end
+
+    #this method will be overriden by plugins who wish to have a defaults params.
+    def self.get_defaults(option)
+      unless HammerCLI::Settings.settings[:defaults].nil? || HammerCLI::Settings.settings[:defaults][option.to_sym].nil?
+        if HammerCLI::Settings.settings[:defaults][option.to_sym][:from_server]
+          value =   Defaults.providers[HammerCLI::Settings.settings[:defaults][option.to_sym][:provider]].get_defaults(option.to_sym)
+          value
```
tstrachota: No need to assign to a variable here and on the line below.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+        new_file.close
+      else
+        DefaultsCommand.file_cant_be_created
+      end
+
+    end
+
+  HammerCLI::MainCommand.subcommand "defaults", _("Defaults management"), HammerCLI::DefaultsCommand
+end
+end
```
tstrachota: Wrong indentation. It would also be good to place the subcommand registration outside the `Defaults` class. I guess it was placed there by mistake as a result of the indentation issue.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+    def self.file_cant_be_created
+      print_message(_("Couldn't add file to ~/.hammer/cli.modules.d please create the path before defaults will be enabled."))
+    end
+
+    def list_all_defaults_message
+      unless HammerCLI::Settings.settings[:defaults].nil?
+        print_message("Hammer defaults list")
+        print_message("--------------------")
+        HammerCLI::Settings.settings[:defaults].each do |key,val|
+          if HammerCLI::Settings.settings[:defaults][key][:from_server]
```
tstrachota: Btw you can do `val[:from_server]`, `val[:value]` and similar everywhere in this block.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+
+    def self.file_cant_be_created
+      print_message(_("Couldn't add file to ~/.hammer/cli.modules.d please create the path before defaults will be enabled."))
+    end
+
+    def list_all_defaults_message
+      unless HammerCLI::Settings.settings[:defaults].nil?
+        print_message("Hammer defaults list")
+        print_message("--------------------")
+        HammerCLI::Settings.settings[:defaults].each do |key,val|
```
tstrachota: It would be good to wrap `HammerCLI::Settings.settings[:defaults]` into some method in the `Defaults` class. For example `Defaults.get_all_defaults` or something similar. This way we can completely hide the usage of settings which will make it easier to make potential future changes.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,143 @@
...
+      option "--plugin-name", "OPTION_PLUGIN_NAME", _("The plugin name defaults will be generated for")
+
+      def execute
+        begin
+          if option_plugin_name.nil? && option_param_val.nil? || !option_plugin_name.nil? && !option_param_val.nil?
+            bad_input
+          else
+            if option_plugin_name
+              namespace = ("HammerCLI" + option_plugin_name.split('_').collect!{ |w| w.downcase.capitalize }.join + "::Defaults").to_sym
+              raise NameError unless HammerCLI::Defaults.providers.keys.include?(namespace)
```
tstrachota: Looking at the namespace, one improvement came to my mind:
What about registering the providers exactly with the same name that is expected in the parameter? I think that would simplify the namespace manipulation. Example:

```ruby
class BaseDefaultsProvider
  def name
    self.class.name.split('::').first.gsub(/^HammerCLI/, '').underscore
    # ...or some other reasonable default
  end
  # not implmented exception for
  # def support?
  # def get_defaults
end

module HammerCLIForeman
  class DefaultsProvider < BaseDefaultsProvider
    # def support?
    # def get_defaults
    # ...
  end
end

HammerCLI::Defaults.register_provider(DefaultsProvider.new)
```

Than you can simply use `HammerCLI::Defaults.providers.key?(provider_name)`. It will also make the `Defaults` class more testable plus we won't restrict the names of the provider classes at all.

tstrachota:
@alongoldboim Thanks for the update! Some notes in the code. Btw it would be good to provide a command for deleting the saved defaults. Users have no other option than editing the defaults.yml manually now.

alongoldboim:
@tstrachota Thanks for all the good notes, the ideas for improvement are really good!
Ill send updates. :)

lib/hammer_cli/defaults.rb
alongoldboim: But we enforce that the path will always be ~/.hammer/cli.modules.d, could just leave it as a plain string?

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,195 @@
+require 'hammer_cli'
+require 'yaml'
+require 'debugger'
```
tstrachota: Tests are failing due to this require...

lib/hammer_cli/defaults.rb
tstrachota: Can plain string, ideally a constant. I want to avoid translators to deal with the implementation specifics.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,195 @@
...
+    def bad_input
+      print_message(_("You must specify value or a plugin name, cant specify both."))
+    end
+
+    def self.file_cant_be_created
+      print_message(_("Couldn't add file to ~/.hammer/cli.modules.d please create the path before defaults will be enabled."))
+    end
+
+    def variable_not_found
+      print_message(_("Couldn't find the requested param in ~/.hammer/cli.modules.d."))
```
tstrachota: % substitution here too, please

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,195 @@
...
+        print_message(_("No defaults file was found"))
+      end
+    end
+
+    autoload_subcommands
+  end
+
+
+  class Defaults
+    PATH = "#{Dir.home}/.hammer/cli.modules.d/defaults.yml"
```
tstrachota: Another candidate for the constant holding `/.hammer/cli.modules.d/`

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,195 @@
...
+    def self.providers
+      @@providers ||= {}
+    end
+
+    def self.defaults_settings
+      @defaults_settings ||= HammerCLI::Settings.settings[:defaults]
+    end
+
+    def self.create_default_file
+      if Dir.exist?("#{Dir.home}/.hammer/cli.modules.d/")
```
tstrachota: Constant here too, please.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,194 @@
+require 'hammer_cli'
+require 'yaml'
+module HammerCLI
+
+  class BaseDefaultsProvider
```
tstrachota: Was there any reason for keeping all methods at class level and avoid creating instances? Maybe I overlooked something but it seems to me that apart from `register_provider` they all can be instance methods.

tstrachota:
@alongoldboim apart from the two notes it looks good to me. Please add tests let's get it merged:)

alongoldboim:
@tstrachota i think there's no reason to instantiate the providers, added some tests, waiting for feedback.

tstrachota:
@alongoldboim I did another round of review and discovered some imperfections. Since I was already in the code I did the review in form of a PR https://github.com/alongoldboim/hammer-cli/pull/1 to keep it on the constructive level. Each commit is one comment with proposed fix.
@mbacovsky please check my proposals too.

tstrachota:
@alongoldboim tests discovered two mistakes I did. Comments above. Can you please fix it. When this is fixed I'm ok to merge. @mbacovsky ?

alongoldboim:
@tstrachota , i've fixed that mistake right now, :) im going over the code making sure everything is right and soon ill squash.

alongoldboim:
@mbacovsky waiting for your green light.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,207 @@
...
+
+    attr_reader :path
+
+    def create_default_file
+      if Dir.exist?(File.dirname(@path))
+        new_file = File.new(path, "w")
+        new_file.write ":defaults:"
+        new_file.close
+      else
+        raise DefaultsPathError.new(_("Couldn't create %s please create the path before defaults will be enabled.") % path)
```
mbacovsky: Sounds "Couldn't create %s please create the path before defaults are enabled."  better?

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,207 @@
...
+    end
+
+    def added_default_message(key, value)
+      print_message(_("Added %{key_val} default-option with value that will be generated from the server.") % {:key_val => key.to_s} ) if value.nil?
+      print_message(_("Added %{key_val} default-option with value %{val_val}.") % {:key_val => key.to_s, :val_val=> value.to_s} ) unless value.nil?
+    end
+
+    def plugin_prob_message
+      print_message(_("Couldn't reach the plugin defaults class, eiter the plugin doesn't have defaults class or the plugin misspelled.(example, for Hammer_CLI_Foreman --plugin-name foreman)"))
+    end
```
mbacovsky: This seems quite difficult to grasp at a first glance. Can it say "Unknown provider. To list available providers run `hammer defaults providers`" and `default providers` would print list of registered providers and supported defaults?

I also see mix of plugin and provider in the messages. It would be simplier to have just defaults providers registered under some name and not imposing implicit connection to plugin that implements the provider.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,207 @@
...
+      end
+    end
+
+    class AddDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'add'
+
+      desc _('Add a default parameter to config')
+      option "--param-name", "OPTION_NAME", _("The name of the default option"), :required => true
+      option "--param-val", "OPTION_VALUE", _("The value for the default option")
+      option "--plugin-name", "OPTION_PLUGIN_NAME", _("The plugin name defaults will be generated for")
```
mbacovsky: --provider-name,  _("The name of the provider providing the value. For list available providers see hammer defaults providers.")

mbacovsky:
The defaults file is getting long, I'd suggest splitting it into separate chunks - base_provider, commands and defaults itself perhaps.

All and all it looks pretty good now. However I still have some usability concerns I stated above (plugins vs provides).

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,207 @@
...
+        HammerCLI::EX_OK
+      end
+    end
+
+    class AddDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'add'
+
+      desc _('Add a default parameter to config')
+      option "--param-name", "OPTION_NAME", _("The name of the default option"), :required => true
+      option "--param-val", "OPTION_VALUE", _("The value for the default option")
```
mbacovsky: `--param-value`

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,207 @@
...
+        end
+        HammerCLI::EX_OK
+      end
+    end
+
+    class AddDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'add'
+
+      desc _('Add a default parameter to config')
+      option "--param-name", "OPTION_NAME", _("The name of the default option"), :required => true
```
mbacovsky: For a first time it was not clear for me if I should use name with dashes or underscores. Example would help:
_("The name of the default option. (e.g. organization_id) ")

alongoldboim:
@mbacovsky good notes!, applying changes.

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,207 @@
...
+        key = key.to_sym
+        defaults[:defaults][key] = value ? {:value => value, :from_server => false} : {:from_server => true, :provider => provider}
+      end
+      write_to_file defaults
+      defaults
+    end
+
+    def get_defaults(option)
+      unless defaults_settings.nil? || defaults_settings[option.to_sym].nil?
+        if defaults_settings[option.to_sym][:from_server]
```
mbacovsky: Isn't the `:from_server` redundant? Is it true that from_server = !!defaults_settings[option.to_sym][:provider].

alongoldboim:
@mbacovsky applied changes.

mbacovsky:
Thanks for the update. I appreciate the tests added. We are almost there. A couple if comments:
- the update files are tab indented while we need spaces
- I like the new providers list, but its usage  should be consistent with the rest of hammer, thus please change it to `hammer defaults list` and `hammer defaults providers`
- provider list doesn't print the supported defaults. I'd suggest to add method `supported_defaults` to the `BaseDefaultsProvider`, if `nil` is returned assume anything is supported. Makes sense?

My version of `defaults providers` for illustration. I don't insist on using `print_collection`, but it provides consistent output and supports export to csv, yaml and others for free (try `hammer --output=yaml defaults providers`):
```ruby
class ProvidersCommand < HammerCLI::DefaultsCommand
  command_name 'providers'
  desc _('List all the providers')

  def execute
    data = context[:defaults].providers.map do |key, val|
      {
        :provider => key.to_s,
        :defaults => (val.supported_defaults || ['*']).map(&:to_s)
      }
    end

    fields = HammerCLI::Output::Dsl.new.build do
      field :provider, _('Provider')
      field :defaults, _('Supported defaults'), Fields::List
    end

    definition = HammerCLI::Output::Definition.new
    definition.append(fields)

    print_collection(definition, data)
    HammerCLI::EX_OK
  end

  def adapter
    @context[:adapter] || :table
  end
end
```

alongoldboim:
@mbacovsky fixed according to comments.

lib/hammer_cli/defaults_commands.rb
```diff
@@ -0,0 +1,163 @@
...
+      end
+    end
+
+    class AddDefaultsCommand < HammerCLI::DefaultsCommand
+      command_name 'add'
+
+      desc _('Add a default parameter to config')
+      option "--param-name", "OPTION_NAME", _("The name of the default option (e.g. organization_id)."), :required => true
+      option "--param-value", "OPTION_VALUE", _("The value for the default option")
+      option "--plugin-name", "OPTION_PLUGIN_NAME", _("The name of the provider providing the value. For list available providers see `hammer defaults list --providers=true`.")
```
mbacovsky: .... see `hammer defaults providers`

mbacovsky: --provider, "OPTION_PROVIDER"

lib/hammer_cli/defaults_commands.rb
```diff
@@ -0,0 +1,163 @@
...
+
+      def execute
+        if option_plugin_name.nil? && option_param_value.nil? || !option_plugin_name.nil? && !option_param_value.nil?
+          bad_input
+          HammerCLI::EX_USAGE
+        else
+          if option_plugin_name
+            namespace = option_plugin_name
+            if !context[:defaults].providers.key?(namespace)
+              plugin_prob_message
```
mbacovsky: print_message(_("Provider #{namespace} was not found. See `hammer defaults providers` for available providers."))

lib/hammer_cli/defaults_commands.rb
```diff
@@ -0,0 +1,163 @@
...
+      print_message(_("Added %{key_val} default-option with value that will be generated from the server.") % {:key_val => key.to_s}) if value.nil?
+      print_message(_("Added %{key_val} default-option with value %{val_val}.") % {:key_val => key.to_s, :val_val => value.to_s}) unless value.nil?
+    end
+
+    def plugin_prob_message
+      print_message(_("Couldn't reach the plugin defaults class, eiter the plugin doesn't have defaults class or the plugin misspelled.(example, for Hammer_CLI_Foreman --plugin-name foreman)"))
+    end
+
+    def defaults_not_supproted_by_plugin
+      print_message(_("The param name is not supported by plugin."))
```
mbacovsky: print_message(_("The param name is not supported by provider. See `hammer defaults providers` for supported params."))

lib/hammer_cli/defaults_commands.rb
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    def param_deleted(param)
+      print_message(_("%{param} was deleted successfully.") % {:param => param.to_s})
+    end
+
+    def bad_input
+      print_message(_("You must specify value or a plugin name, cant specify both."))
```
mbacovsky: *provider name

lib/hammer_cli/defaults_commands.rb
```diff
@@ -0,0 +1,163 @@
...
+    def added_default_message(key, value)
+      print_message(_("Added %{key_val} default-option with value that will be generated from the server.") % {:key_val => key.to_s}) if value.nil?
+      print_message(_("Added %{key_val} default-option with value %{val_val}.") % {:key_val => key.to_s, :val_val => value.to_s}) unless value.nil?
+    end
+
+    def plugin_prob_message
+      print_message(_("Couldn't reach the plugin defaults class, eiter the plugin doesn't have defaults class or the plugin misspelled.(example, for Hammer_CLI_Foreman --plugin-name foreman)"))
+    end
+
+    def defaults_not_supproted_by_plugin
```
mbacovsky: *supported, but I'd prefer
```ruby
def defaults_not_supported_by_provider
```

lib/hammer_cli/defaults.rb
```diff
@@ -0,0 +1,79 @@
+require 'hammer_cli/defaults_commands'
+module HammerCLI
+  class Defaults
+    DEFAULT_FILE = "#{Dir.home}/.hammer/cli.modules.d/defaults.yml"
```
mbacovsky: `cli.modules.d` is for hammer module configuration. This config should be in "#{Dir.home}/.hammer/defaults.yml". What dou you think? @tstrachota?

tstrachota: Yes, it's probably better place.

mbacovsky:
I like the new list commands, thanks! In the comments above I pointed out some UX concerns. Besides that it is good to go for me.

mbacovsky:
Btw how do we want to handle nested params? Let's say I have `organization_id` set to '5'. 
If I create new host should it be created within organization with id 5 by default or not? 
It is not now as I tried it. Is it a bug or expected behavior? @alongoldboim, @tstrachota ?

tstrachota:
@mbacovsky imho it should assign host to that organization. It should behave the same as when you switch context in UI.

alongoldboim:
@mbacovsky I agree with @tstrachota, ill check the reason it's not behaving like that now.

alongoldboim:
@mbacovsky @tstrachota Fixed all the above.
Regarding the Host creation, the cmd doesn't work without specifying organization/location, so the defaults won't work anyway, if we decide to take this validation off it will expose, i think for commands that require some parameter we won't be able to insert the defaults values (taking into mind that if someone mention a param on the cli we wont override that).

alongoldboim:
@mbacovsky @tstrachota tests are failing, ill fix them, but in general i changed the place we insert our defaults, are you ok with the change?

alongoldboim:
@mbacovsky @tstrachota changed a bit of the logic on when the defaults are added, and changed the tests accordingly, please give it a check.

lib/hammer_cli/abstract.rb
```diff
@@ -233,10 +232,12 @@ def self.option(switches, type, description, opts = {}, &block)
     end
 
     def all_options
-      self.class.recognised_options.inject({}) do |h, opt|
-        h[opt.attribute_name] = send(opt.read_method)
-        h
+      self.class.recognised_options.inject({}) do |hash, opt|
+        hash[opt.attribute_name] = send(opt.read_method)
+        hash[opt.attribute_name] = context[:defaults].get_defaults(opt.attribute_name) if hash[opt.attribute_name].nil? && context[:defaults]
```
mbacovsky: Can you keep the logging of the default being used that was present in previous version. It can make bug hunting easier in the future.

mbacovsky:
@alongoldboim, I like the new location of getting the defaults. It seems more flexible as you can defaults for options that are not comming from API. Well played. It also works fine and the user experience is good.

When the logging is added I'm for merging it. :trumpet:

alongoldboim:
@mbacovsky yea that was part of the idea in moving it there :), thanks, will add logging.

alongoldboim:
@mbacovsky @tstrachota  added some fixes to improve performance.

lib/hammer_cli/abstract.rb
```diff
@@ -233,10 +232,16 @@ def self.option(switches, type, description, opts = {}, &block)
...
-        h
+      @all_options ||= self.class.recognised_options.inject({}) do |hash, opt|
+        hash[opt.attribute_name] = send(opt.read_method)
+        hash[opt.attribute_name] = add_custom_defaults(opt.attribute_name) if hash[opt.attribute_name].nil?
+        hash
       end
+      @all_options
+    end
+
+    def add_custom_defaults(attr)
```
mbacovsky: I'm still missing the logging of the defaults being used. The `add_custom_defaults` is good candidate for doing that:
```ruby
    def add_custom_defaults(attr)
      if context[:defaults]
        value = context[:defaults].get_defaults(attr)
        logger.info("Custom default value #{value} was used for attribute #{attr}") if value
      end
      value
    end
```
I'd also make it `private`, as it is not intended to be part of public command interface

alongoldboim:
@mbacovsky logger added, moved to private.

mbacovsky:
@alongoldboim, thanks! I still need to test it, but it looks good now.

mbacovsky:
@alongoldboim It looks good and works as expected now! Thank you. :beers: :construction_worker: 
:+1: form me. @tstrachota could you confirm it is okay to merge?

alongoldboim:
:+1:

test/functional/defaults_test.rb
```diff
@@ -0,0 +1,206 @@
...
+      @defaults.stubs(:defaults_settings).returns({})
+
+      out, err = run_cmd(HammerCLI::DefaultsCommand::ListDefaultsCommand, [])
+      assert_equal "", err
+      assert_equal expected_message(header), out
+    end
+  end
+
+  describe 'defaults providers' do
+	  header = ['---------|-------------------',
```
tstrachota: Tab indentation

test/functional/defaults_test.rb
```diff
@@ -0,0 +1,206 @@
...
+              'PROVIDER | SUPPORTED DEFAULTS',
+              '---------|-------------------',
+              ''
+	  ]
+
+    it 'prints all providers and their supported defaults' do
+	    data = ['foreman  | organization_id   ',
+	            '---------|-------------------',
+	            ''
+	    ]
```
tstrachota: Same here, the whole block above is tab indented.

lib/hammer_cli/settings.rb
```diff
@@ -1,6 +1,5 @@
 require 'yaml'
 require 'logging'
-
```
tstrachota: I'd limit whitespace changes only to the necessary ones.

tstrachota:
@alongoldboim :+1: pending the whitespace issues. Please set your editor to use spaces instead of tabs.

alongoldboim:
@tstrachota besides those, anything serious to change?

alongoldboim:
@tstrachota @mbacovsky done deal.

mbacovsky:
@alongoldboim Thanks, merging!!! :cocktail:

PR #176, opened 2015-08-09
refs #11280 - drop Ruby 1.8 support

mmoll:
I did leave rest-client untouched for now, as 1.7 does verify_ssl by default.
I'm unsure about the major version bump of logging, but at least an error got logged as expected with 2.0.0 in my tests.

Discuss! :boom:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,17 +2,14 @@ source "https://rubygems.org"
 
 gemspec
 
-if RUBY_VERSION >= '1.9'
-  # for generating i18n files, gettext > 3.0 dropped ruby 1.8 support
-  gem 'gettext', '>= 3.1.3', '< 4.0.0'
-end
+gem 'gettext', '>= 3.1.3', '< 4.0.0'
 
 group :test do
   gem 'rake', '~> 10.1.0'
   gem 'thor'
   gem 'minitest', '4.7.4'
   gem 'minitest-spec-context'
-  gem 'simplecov', '< 0.9.0' # 0.9.0 is not compatible with Ruby 1.8.x
+  gem 'simplecov'
   gem 'mocha'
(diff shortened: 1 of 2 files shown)


mmoll:
[test]

domcleal:
We're still a bit limited on what we can update with regards to RPM installations, since some dependencies are shared by kafo and the smart proxy, which still support 1.8.7.  We don't have the means to build multiple versions of one package (I'm unsure if this is technically possible with our packaging repo, needs investigation).

This would affect logging, highline and rest-client among others.  It'd be fine to lift the maximum version in hammer-cli itself, provided the minimum isn't raised!  We'd need to ensure logging 1.x, highline 1.6 and rest-client 1.6 etc also work.

mmoll:
Yes, at least with highline I'm very sure that it works with <1.7 on Ruby 1.8 and 1.7.x on higher Ruby versions. With logging I'm unsure, as said and with rest-client the problem is more related to the verify_ssl thing which I want to tackle seperately.

In general I can say we'll run into a lot of problems when we will start pckaging for Ubuntu 16.04, as quite some gem packages already where updated in upstream Debian/Ubuntu now.

domcleal:
I think for logging it was only a small incompatibility that we hit in Foreman core, and decided to support both for the same reasons: https://github.com/theforeman/foreman/blob/develop/lib/foreman/logging.rb#L96-L101

But yep, fine by me to raise maximums wherever possible.

mmoll:
@mbacovsky @tstrachota :point_up:

tstrachota:
I did some testing and it works ok. So fine by me to merge that. @mbacovsky ?

mbacovsky:
:+1:  While rest-client dependency will be fixed in separate PR this looks good to merge.

tstrachota:
Merging, thanks @mmoll

PR #177, opened 2015-08-18
Fixes #11158 - abort when custom config file is not found

tstrachota:


mbacovsky:
Works well. I would probably prefer to abort, because if the config you explicitly asked for is missing it won't probably work anyway.

mbacovsky:
Could you please fix the commit message as well?

mbacovsky:
:metal: LGTM

PR #178, opened 2015-08-20
Fixes #11439 - Prevents hammer from reading './config' file

adamruzicka:


bin/hammer
```diff
@@ -52,12 +52,16 @@ end
 
 CFG_PATH_LOCAL = ['./config/']
 
+HammerCLI::Settings.load_from_paths CFG_PATH_LOCAL
+
 if preparser.config
-  CFG_PATH_LOCAL.unshift preparser.config
+  if File.file? preparser.config
+    HammerCLI::Settings.load_from_file preparser.config
```
tstrachota: Does it work with tilde too? It seems that `load_from_file` does not expand paths.

adamruzicka: The tilde shouldn't even get there. It should get expanded by the shell, unless the user escapes it, in which case I'd think the tilde is part of the path as a literal

tstrachota: That's not completely true. Bash won't expand it when you pass the parameter as a string. E.g.
```
hammer -c '~/some_config.yml'
```
I think this should be supported usage.

tstrachota:
thanks @adamruzicka

PR #179, opened 2015-08-20
Fixes #7534 - Add filtering of the logs

mbacovsky:
Ability to filter the data with simple gsub was added in LogEvent (Logging internal object representing the data being logged). LogEvent class can hold list of `pattern`, `replacement` pairs that are applied one by one. The list can be extended with 
```ruby
Logging::LogEvent.add_data_filter(/foo/, 'bar')
```

Existing filtering of password arguments was replaced.
Tests were updated accordingly.

### DIFF
--- a/lib/hammer_cli/abstract.rb
+++ b/lib/hammer_cli/abstract.rb
@@ -30,9 +30,7 @@ module HammerCLI
     def parse(arguments)
       super
       validate_options
-      safe_options = options.dup
-      safe_options.keys.each { |k| safe_options[k] = '***' if k.end_with?('password') }
-      logger.info "Called with options: %s" % safe_options.inspect
+      logger.info "Called with options: %s" % options.inspect
     rescue HammerCLI::Validator::ValidationError => e
       signal_usage_error e.message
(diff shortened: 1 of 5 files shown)


tstrachota:
It fixes the problem from what I've been able to test so far.

@mbacovsky did you consider creating custom layout for our logging? It seems to be more natural place for such code. It would also allow us to avoid opening the `LogEvent` class and scan data other than string too. `Logging::Layouts::Pattern` which we use now seems to be meta-programming hell so I'm not sure what the effort would be and whether it's worth it.

mbacovsky:
@tstrachota I like your idea. It seems the implementation would be more complex then current solution. Not sure if it is worth the effort.

tstrachota:
@mbacovsky ok, let's keep your solution. Please add the removed trailing lines to the files (seems like wrong editor setting) and we can merge it.

mbacovsky:
White space fixed. Thanks for review @tstrachota

tstrachota:
Merged, thanks @mbacovsky

PR #181, opened 2015-09-11
Bkearney/new downstream strings

bkearney:


bkearney:
Since the new and changed one was merged, this one is not needed.

PR #186, opened 2015-12-08
Add Catalan language

rbuj:


tstrachota:
Similar as in https://github.com/theforeman/hammer-cli-foreman/pull/207 I changed `Id.` to `ID` and merged as ff0ab5016791ef1f93e87a9d92114063aad6857b Thanks @rbuj

PR #187, opened 2015-12-09
Fixes #12751 - --plugin-name renamed to --provider

mbacovsky:
- renamed option in `defaults add`
- fixed minor formatting issues
- fixed plugin/provider inconsistencies in the code
- removed default provider_name

I propose removal of the provider default name generated from the plugin name. The reason is that there is no relation between the provider name  and the plugin name. I'd prefer one plugin to offer rather more simplier defaults providers then one complex provider. E.g. foreman plugin could implement `taxonomies` provider providing orgs and locations. 

It would be nice to add description to the provider and print it in `defaults providers` so that the user knows what is it good for. What do you think? @alongoldboim, @tstrachota

### DIFF
--- a/lib/hammer_cli/defaults.rb
+++ b/lib/hammer_cli/defaults.rb
@@ -15,7 +15,7 @@ module HammerCLI
     end
 
     def register_provider(provider)
-      providers[provider.plugin_name.to_s] = provider
+      providers[provider.provider_name.to_s] = provider
     end
 
     def providers
--- a/lib/hammer_cli/defaults_commands.rb
+++ b/lib/hammer_cli/defaults_commands.rb
@@ -2,23 +2,19 @@ require 'hammer_cli'
 require 'yaml'
 module HammerCLI
   class BaseDefaultsProvider
-    def self.plugin_name
-      self.name.split('::').first.gsub(/^HammerCLI/, '').underscore
+    attr_reader :provider_name, :supported_defaults, :description
+
+    def initialize
+      @provider_name = nil
+      @supported_defaults = nil
+      @description = 'Abstract provider'
     end
 
-    def self.register_provider
-      HammerCLI.defaults.register_provider(self)
+    def param_supported?(param)
+      @supported_defaults.nil? || @supported_defaults.any? {|s| s.to_s == param}
     end
 
-    def self.support?
-      raise NotImplementedError
-    end
-
-    def self.supported_defaults
-      raise NotImplementedError
-    end
-
-    def self.get_defaults
+    def get_defaults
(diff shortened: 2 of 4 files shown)


alongoldboim:
@mbacovsky @tstrachota I think those changes are good, sticking with 1 name convention to reduce confusion (provider vs plugin).
Placing the name of the provider in the provider hands is also good, will be more flexible.
:+1:

mbacovsky:
In the second commit I've added some thoughts on refactoring. I changed Providers to be used as an instance, added description and renamed method `support?`.

The foreman provider would look like this:
```ruby
require 'hammer_cli'
module HammerCLIForeman
  class TaxonomiesDefaults < HammerCLI::BaseDefaultsProvider
    def initialize
      @provider_name = 'taxonomies'
      @supported_defaults = [:organization_id, :location_id]
      @description = _('Use the default organization and/or location from the server')
    end

    def get_defaults(param)
      param = "default_organization" if param == :organization_id
      param = "default_location" if param == :location_id
      user = HammerCLIForeman.foreman_resource(:users).action(:index).call(:search =>"login="+HammerCLIForeman.credentials.username)
      val = nil
      if user
        val = user["results"].first[param] if user["results"]
        val = val["id"] if val.is_a?(Hash) && param.include?("default")
      end
      val
    end
  end
  HammerCLI.defaults.register_provider(TaxonomiesDefaults.new())
end
```
@alongoldboim, please take a look if this is something we want in.

alongoldboim:
@mbacovsky it looks good to me and i like that we are using instance variables instead of methods so BaseDefaultsProvider is clearer, but i still think keeping the provider name as foreman is more user friendly but i have been mistaken before... :)
Lets wait and see what @tstrachota has to say about this change.

tstrachota:
* +1 for the plugin/provider unification
* big +1 for registering the providers as instances
* I like throwing `NotImplemented`exceptions more but no strong opinion here, it's a cosmetic thing
* Giving more freedom in provider names is probably good.
* Calling the provider `taxonomy` would enable for creating more providers for foreman from the start, but: Do we use the name `taxonomies` anywhere else in UI/cli/api? I think it's org/loc everywhere so it could be a bit confusing for new users. Sticking with `"foreman"` and re-evaluating later if needed is fine by me. It's not a big issue.

mbacovsky:
Squashed

tstrachota:
I forgot to mention I like the provider description too. Good idea!
Thanks @mbacovsky @alongoldboim, merging

alongoldboim:
btw changing the tests to work with regex is a lot better then the previous version :+1:

PR #192, opened 2016-02-02
Fixes #13525: Allows loading module configs from config/

ehelms:
Hammer has an expecatation that module configuration exists in
cli.modules.d and typically from the system configuration directory
or user configuration directory. If a user wants to install hammer
or a module via 'gem install' the modules aren't available by default.
Most modules put their default configuration into <module>/config/<module>.yml
which won't be loaded by default. This change allows gem install of
any module and automatic usage of hammer and any modules that are
enabled by default.

### DIFF
--- a/bin/hammer
+++ b/bin/hammer
@@ -42,6 +42,13 @@ end
 # load user's settings
 require 'hammer_cli/settings'
 
+hammer_gems = Gem::Specification.select do |spec|
+  spec.name.include?('hammer_cli')
+end
+CFG_PATH_GEM = hammer_gems.collect { |spec| "#{spec.gem_dir}/config" }
+
+HammerCLI::Settings.load_from_paths CFG_PATH_GEM
+
 CFG_PATH_LEGACY = ['~/.foreman/', '/etc/foreman/', "#{::RbConfig::CONFIG['sysconfdir']}/foreman/"].uniq
 HammerCLI::Settings.load_from_paths CFG_PATH_LEGACY
 if HammerCLI::Settings.path_history.empty?
--- a/lib/hammer_cli/settings.rb
+++ b/lib/hammer_cli/settings.rb
@@ -20,7 +20,7 @@ module HammerCLI
           load_from_file(File.join(full_path, 'cli_config.yml'))
           load_from_file(File.join(full_path, 'defaults.yml'))
           # load config for modules
-          Dir.glob(File.join(full_path, 'cli.modules.d/*.yml')).sort.each do |f|
+          Dir.glob(File.join(full_path, '**/*.yml')).sort.each do |f|
             load_from_file(f)
           end
           Dir.glob(File.join(full_path, 'hammer.modules.d/*.yml')).sort.each do |f|


ehelms:
I would consider keeping the 'cli.modules.d' paradigm here, but it will require updating each CLI module to move the default config to `config/cli.modules.d/<module>.yml` and releasing a new version of each.

tstrachota:
@ehelms thanks for bringing this up. I feel it's a problem too, but I'm afraid of loading the defaults blindly on each hammer start.

I've been thinking (I'm brainstorming now) about some utility that would become a standard way of enabling plugins. Something like `hammer-plugin list` to list available plugins and `hammer-plugin enable` that would copy the files into user's conf dir. Not sure how we would do `disable` though, since editing the config files from a script is not a good idea (loosing comments and formatting).

@ehelms @mbacovsky What do you think?

lib/hammer_cli/settings.rb
```diff
@@ -20,7 +20,7 @@ def self.load_from_paths(files)
           load_from_file(File.join(full_path, 'cli_config.yml'))
           load_from_file(File.join(full_path, 'defaults.yml'))
           # load config for modules
-          Dir.glob(File.join(full_path, 'cli.modules.d/*.yml')).sort.each do |f|
+          Dir.glob(File.join(full_path, '**/*.yml')).sort.each do |f|
```
tstrachota: This may potentially break plugins that have more yaml files than just the standard one. Not sure how it would work for example with `hammer-cli-import` https://github.com/Katello/hammer-cli-import/tree/master/config

ehelms:
What is the fear of loading defaults blindly on start? Or loading them as the baseline, and any user configurations set override that?  I wouldn't mind an enable/disable style, however, I am not seeing a situation where I would take the time to install the plugin and not want it to be enabled and ready to go?

For plugins that have their own configurations and break from convention, well, we could enforce a convention? That may requires updates to some of these plugins, but I would think it not too much work and the benefit would out weight the cost.

komidore64:
I'm in favor of Hammer enabling any discovered plugins by default. The user can disable them for some reason if need be via an option.

tstrachota:
Ok, I checked all hammer plugins I found on github and the only one that could potentially break seems to be `hammer_cli_import`. So if we warn the developers in advance and make sure it doesn't break I'm ok with merging it.
@ehelms would you mind testing your change with the import plugin?

bin/hammer
```diff
@@ -42,6 +42,13 @@ end
...
+hammer_gems = Gem::Specification.select do |spec|
+  spec.name.include?('hammer_cli')
+end
+CFG_PATH_GEM = hammer_gems.collect { |spec| "#{spec.gem_dir}/config" }
+
+HammerCLI::Settings.load_from_paths CFG_PATH_GEM
+
 CFG_PATH_LEGACY = ['~/.foreman/', '/etc/foreman/', "#{::RbConfig::CONFIG['sysconfdir']}/foreman/"].uniq
 HammerCLI::Settings.load_from_paths CFG_PATH_LEGACY
 if HammerCLI::Settings.path_history.empty?
```
tstrachota: This condition causes failures for me. When there are some default configs discovered and loaded, the path is recorded in the path history. The condition was there to check for legacy config files. As a result the user's configuration is skipped:
```
hammer 
Warning: Legacy config paths detected, move the following files
    /root/hammer/hammer-cli-foreman/config/foreman.yml -> /root/hammer/hammer-cli-hammer/config/foreman.yml
    /root/hammer/hammer-cli/config/cli_config.template.yml -> /root/hammer/hammer-cli/config/cli_config.template.yml
Could not load the API description from the server
  - is the server down?
  - was 'foreman-rake apipie:cache' run on the server when using apipie cache? (typical production settings)
Warning: An error occured while loading module hammer_cli_foreman
```

I think we can remove the legacy code now. It's been there for long enough: https://github.com/theforeman/hammer-cli/commit/9959a4fb145d1970cc1e70abba0023d177b82a6f

lib/hammer_cli/settings.rb
lzap: It's a great chance to change the `load_from_paths` method to also accept non-arrays (e.g. string containing just one path). Can you do this?

mmoll:
What's there status here?

ofedoren:
@ehelms. ping.

shiramax:
I'm closing this issue at the moment because the current implementation needs to be discussed further.

PR #193, opened 2016-02-12
Fixes #12797 - remove psych require from Gemfile

tstrachota:
It was used only as a workaroud for a bundler issue that has been fixed.


mbacovsky:
:+1: Thanks @tstrachota

PR #194, opened 2016-02-19
Fixes #4118 - support for command testing moved to core

tstrachota:
Support for testing that can be shared with plugins has been moved from hammer-cli-foreman to core.
Functional test for defaults have been modified to use the new libraries.

See also https://github.com/Katello/hammer-cli-katello/pull/353 for usage demonstration.


lib/hammer_cli/testing/output_matchers.rb
```diff
@@ -0,0 +1,113 @@
...
+          Regexp.new(Regexp.quote(label) + ':[ ]+' + Regexp.quote(value))
+        end
+      end
+
+      class OutputMatcher < MatcherBase
+        attr_accessor :expected_lines
+
+        def initialize(expected="", options={})
+          @expected_lines = expected.is_a?(Array) ? expected : [expected]
+          @ignore_whitespace = options[:ignore_whitespace].nil? ? true : options[:ignore_whitespace]
```
mbacovsky: nitpick: `options.fetch(:ignore_whitespace, true)`

lib/hammer_cli/testing/output_matchers.rb
```diff
@@ -0,0 +1,113 @@
...
+          assert(actual.include?(expected_lines), message)
+        end
+
+        protected
+
+        def striped_lines
+          @expected_lines.map do |line|
+            line.gsub(/\s*$/, '')
+          end
+        end
```
mbacovsky: Could be more DRY and I assume rstrip is more efficient:
```ruby
def strip_lines(lines)
  lines.map(&:rstrip)
end
```
and reuse in `strip_lines(@expected_lines)` and `strip_lines(actual_split("\n")).join("\n")`

tstrachota:
Thanks @mbacovsky, updated.

mbacovsky:
I like the making the test tools available to plugin developers. While making it public it would deserve some documentation or guidelines suggesting how to use it, could you make a Doc RFE for it?
+:100: (ACK) from me, thanks @tstrachota!

tstrachota:
@mbacovsky RFE issue for the docs: http://projects.theforeman.org/issues/13851

daviddavis:
:+1:

PR #197, opened 2016-03-04
Fixes #12869 - enable json for key=value parameters

tstrachota:
Capabilities of the current format are quite limited. It allows only for entering either --parameters='key1=value1,key2=value2' or --parameters='key1=[val1,val2,val3]'. It's also not very intuitive, so users often pass incorrect values.

This PR adds support for json format with backward compatibility with the current one. Parameters can now be then entered as both:
`--parameters='key1=value1'`
or
`--parameters='{"key1": "value1", "nested": {"array": [1, 2]}}'`

### DIFF
--- a/lib/hammer_cli/options/normalizers.rb
+++ b/lib/hammer_cli/options/normalizers.rb
@@ -22,6 +22,9 @@ module HammerCLI
 
       class KeyValueList < AbstractNormalizer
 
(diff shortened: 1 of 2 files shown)


mbacovsky:
:+1: Thanks @tstrachota! :ice_cream:

PR #198, opened 2016-03-08
Fixes #14092 -  gettext speed improvements

tstrachota:
Fixes fast-gettext performance issues in multidomain mode:
* multiple domains replaced by a translation repository that merges subrepositories
* .mo files are loaded lazily
* translation api for plugins remains the same

Please compare hammer speed with and without this patch when it's configured with multiple plugins.

### DIFF
--- a/lib/hammer_cli/i18n.rb
+++ b/lib/hammer_cli/i18n.rb
@@ -4,68 +4,40 @@ require 'locale'
 module HammerCLI
   module I18n
 
-    module AllDomains
-      def _(key)
-        FastGettext::TranslationMultidomain.D_(key)
-      end
-
-      def n_(*keys)
-        FastGettext::TranslationMultidomain.Dn_(*keys)
-      end
-
-      def s_(key, separator=nil)
-        FastGettext::TranslationMultidomain.Ds_(key, separator)
-      end
-
-      def ns_(*keys)
-        FastGettext::TranslationMultidomain.Dns_(*keys)
-      end
-    end
+    TEXT_DOMAIN = 'hammer-cli'
 
     # include this module to see translations highlighted
     module Debug
       DL = '>'
       DR = '<'
 
-      # slightly modified copy of fast_gettext D_* method
+      # slightly modified copy of fast_gettext _ method
       def _(key)
-        FastGettext.translation_repositories.each_key do |domain|
-          result = FastGettext::TranslationMultidomain.d_(domain, key) {nil}
-          return DL + result + DR unless result.nil?
-        end
-        DL + key + DR
+        _wrap { FastGettext::Translation._(key) }
       end
 
-      # slightly modified copy of fast_gettext D_* method
+      # slightly modified copy of fast_gettext n_ method
       def n_(*keys)
-        FastGettext.translation_repositories.each_key do |domain|
-          result = FastGettext::TranslationMultidomain.dn_(domain, *keys) {nil}
-          return DL + result + DR unless result.nil?
-        end
-        DL + keys[-3].split(keys[-2]||FastGettext::NAMESPACE_SEPARATOR).last + DR
+        _wrap { FastGettext::Translation.n_(*keys) }
       end
 
-      # slightly modified copy of fast_gettext D_* method
-      def s_(key, separator=nil)
-        FastGettext.translation_repositories.each_key do |domain|
-          result = FastGettext::TranslationMultidomain.ds_(domain, key, separator) {nil}
-          return DL + result + DR unless result.nil?
-        end
-        DL + key.split(separator||FastGettext::NAMESPACE_SEPARATOR).last + DR
+      # slightly modified copy of fast_gettext s_ method
+      def s_(key, separator=nil, &block)
+        _wrap { FastGettext::Translation.s_(key, separator, &block) }
       end
 
-      # slightly modified copy of fast_gettext D_* method
-      def ns_(*keys)
-        FastGettext.translation_repositories.each_key do |domain|
-          result = FastGettext::TranslationMultidomain.dns_(domain, *keys) {nil}
-          return DL + result + DR unless result.nil?
-        end
-        DL + keys[-2].split(FastGettext::NAMESPACE_SEPARATOR).last + DR
+      # slightly modified copy of fast_gettext ns_ method
+      def ns_(*args, &block)
+        _wrap { FastGettext::Translation.ns_(*args, &block) }
+      end
+
+      def _wrap(&block)
+        result = yield
+        DL + result + DR unless result.nil?
       end
     end
 
     class AbstractLocaleDomain
-
       def available_locales
         Dir.glob(locale_dir+'/*').select { |f| File.directory? f }.map { |f| File.basename(f) }
       end
@@ -87,7 +59,6 @@ module HammerCLI
 
 
     class LocaleDomain < AbstractLocaleDomain
-
       def translated_files
         Dir.glob(File.join(File.dirname(__FILE__), '../**/*.rb'))
       end
@@ -99,17 +70,13 @@ module HammerCLI
       def domain_name
         'hammer-cli'
       end
-
     end
 
     class SystemLocaleDomain < LocaleDomain
-
       def locale_dir
         '/usr/share/locale'
       end
-
     end
-
 
     def self.locale
       lang_variant = Locale.current.to_simple.to_str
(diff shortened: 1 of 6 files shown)


domcleal:
Is there anything here that could be sent as a patch to fast_gettext instead?  We use the same gem in Foreman itself, so it'd be great to get any performance enhancements you've added here in Foreman (and other apps) too, without copying & pasting.

tstrachota:
@domcleal they have an issue open for it https://github.com/grosser/fast_gettext/issues/75 I plan to send a PR at least with the lazy mo files. The unified repository could be moved to fast-gettext too.
I'd keep the code in hammer-cli for now and refactor when the patches into fast-gettext are accepted.

kbrock:
@tstrachota any success getting this into PR over in fast_gettext land?

tstrachota:
@kbrock I haven't found time for it yet, but it's still on my todo.

lib/hammer_cli/i18n.rb
```diff
@@ -99,17 +70,111 @@ def locale_dir
...
 
     class SystemLocaleDomain < LocaleDomain
-
       def locale_dir
         '/usr/share/locale'
       end
+    end
+
+    class LazyMoFile < FastGettext::MoFile
+      attr_reader :data
```
mbacovsky: attr_reader overriden by method `data` below

lib/hammer_cli/i18n.rb
```diff
@@ -99,17 +70,111 @@ def locale_dir
...
+        else
+          @data = FastGettext::GetText::MOFile.open(file, "UTF-8")
+        end
+        make_singular_and_plural_available
+      end
 
+      def [](key)
+        load_data(@filename) if @data.nil?
+        @data[key]
+      end
```
mbacovsky: `data[key]` would be more DRY

mbacovsky:
It looks good overall. I measured about 15-20% shorter load times in hammer which is great improvement.
Besides minor things it would be good to merge but lets wait on if fast_gettext accepts your PR. :poodle: :zzz:

jrafanie:
> Besides minor things it would be good to merge but lets wait on if fast_gettext accepts your PR

@tstrachota @mbacovsky was there a PR I could review/help get merged ❤️ 👏 👍 🙇 

We are getting crushed with loading multiple message catalogs ManageIQ whenever we load rails (at least 15% of load time is loading 4 or 5 catalogs).  I am happy to try out any WIP branch you may have.  🙏 

Thanks in advance for sharing your optimization!

tstrachota:
@jrafanie @mbacovsky here it is https://github.com/grosser/fast_gettext/pull/79

tstrachota:
Second part of extracting the code into fast_gettext is in progress:
https://github.com/grosser/fast_gettext/pull/82

I updated the PR accordingly, you can test it with the fast_gettext's branch now.

tstrachota:
@mbacovsky the changes in fast_gettext have been merged and 1.2.0 was released. I consider the PR to be final now. You can re-test.

tstrachota:
[test]

akofink:
What's the plan here? Are we dropping support for Ruby 2.0.0?

tstrachota:
Good catch. I added compatibility code that will allow us to run with older versions of fast_gettext. The merge repo speed improvement will simply not apply and a fallback chain repo will be used at systems with ruby 2.0. From those we provide packages for it should be only Ubuntu Trusty. Other systems either have ruby 2.1+ or hammer runs in scl.

akofink:
Wouldn't we need to add some logic to the gemspec/Gemfile to install an older version (<= 1.1.0) of fast_gettext in Ruby 2.0 due to [this](https://github.com/grosser/fast_gettext/blob/master/fast_gettext.gemspec#L11) Ruby version requirement since fast_gettext v. >= 1.2.0?

tstrachota:
Probably the only think we can do is to conditionally lock the version in the Gemfile
```ruby
if RUBY_VERSION < '2.1.0'
  gem 'fast_gettext', '< 1.2.0'
end
```
I don't think there's a solution for a direct gem installation (other than releasing a special gem for ruby2.0 which I'd like to avoid). `rpm` and `deb` can solve it on per OS basis.

mbacovsky:
@tstrachota, the code looks nice and clean. I appreciate the tests added. 

Regarding the version issues (gettext 1.2.0 on ruby 2.0.0) - I find the suggested solution acceptable and as after reasonable time nobody on the lists had no objections I'd go with it.

What confuses me is a performance during my tests. I tried with your patch and without cross with fast_gettext 1.0.0 and 1.2.0. My ruby version is 2.2.5. To measure the impact I ran `hammer --help` as there is hammer loading involved and decent amount of translating. I measured 20 invocations in a loop. 

```
$ git st|head -n1
On branch master

$ bundle |grep fast_gettext
Using fast_gettext 1.2.0

$ time for i in {1..20}; do hm -s "https://192.168.121.243" -h > /dev/null; done
real	0m53.098s
user	0m48.907s
sys	0m4.121s

$ bundle |grep fast_gettext
Using fast_gettext 1.0.0

$ time for i in {1..20}; do hm -s "https://192.168.121.243" -h > /dev/null; done
real	0m54.275s
user	0m49.952s
sys	0m4.274s

$ git st|head -n1
HEAD detached at tstrachota/gettext_prototype

$ bundle |grep fast_gettext
Using fast_gettext 1.2.0

$ time for i in {1..20}; do hm -s "https://192.168.121.243" -h > /dev/null; done
real	0m47.502s
user	0m43.551s
sys	0m3.879s

$ bundle |grep fast_gettext
Using fast_gettext 1.0.0

$ time for i in {1..20}; do hm -s "https://192.168.121.243" -h > /dev/null; done
real	0m48.212s
user	0m44.036s
sys	0m4.134s
```
As you can see in the output there was about 12% better performance with your patch, regardless on `fast_gettext` version. I checked that hammer properly used the `merge` strategy with 1.2.0 (`chain` otherwise) but the difference was about 1%. Two questions: 
- why there is almost no impact of fast_gettext 1.2.0 on performance? 
- why the performance boost now is only 12% vs. 15-20% measured on the original PR (I must admit the metric was not the same so the difference could be irrelevant) ?

Do you see any issue with how I measured the performance and does it match your tests?

ACK from me as soon as the concerns about performance are clear.

tstrachota:
@mbacovsky the impact on the performance very much depends on the number of plugins you use in hammer. More plugins you have the more noticeable improvement should be.

Impact of using the merge repo is highest on commands from plugins that use a lot of translated strings. Running `host create -h` with at least 4 other translated plugins enabled would be a good candidate to test that.

In short: Lazy loading improves startup speed, Merge repo improves speed of translation lookup.

mbacovsky:
My hammer configuration was 
```
$ hm -s "https://192.168.121.243" --version
hammer (0.7.0)
 * hammer_cli_console (0.0.1)
 * hammer_cli_foreman (0.7.0)
 * hammer_cli_foreman_docker (0.0.3)
 * hammer_cli_katello (0.0.26)
```
and it was same is in my original performance testing. I'll retry with hammer host create --help if it shows any difference with 1.2.0

mbacovsky:
I used for the test `host create -h` results are attached. The difference between master with `fast_gettext = 1.0.0` and your branch with `fast_gettext = 1.2.0` is about 13%. Almost no difference between `1.0.0` and `1.2.0` on your branch does not feel right. 

```
$ git st|head -n1
On branch master

$ bundle |grep fast_gettext
Using fast_gettext 1.2.0

$ time for i in {1..20}; do hm -s "https://192.168.121.243" host create -h > /dev/null; done
real	0m56.941s
user	0m52.443s
sys	0m4.441s

$ bundle |grep fast_gettext
Using fast_gettext 1.0.0

$ time for i in {1..20}; do hm -s "https://192.168.121.243" host create -h > /dev/null; done
real	0m59.223s
user	0m54.562s
sys	0m4.585s

$ git st|head -n1
HEAD detached at tstrachota/gettext_prototype

$ bundle |grep fast_gettext
Using fast_gettext 1.2.0

$ time for i in {1..20}; do hm -s "https://192.168.121.243" host create -h > /dev/null; done
real	0m52.133s
user	0m47.702s
sys	0m4.370s

$ bundle |grep fast_gettext
Using fast_gettext 1.0.0

$ time for i in {1..20}; do hm -s "https://192.168.121.243" host create -h > /dev/null; done
real	0m52.896s
user	0m48.204s
sys	0m4.495s
```

mbacovsky:
Hovever I didn't found any issue and I tested the merge startegy is used when expected.
ACK form me.

mbacovsky:
Based on discussion and further testing done by @tstrachota I'm going to merge this.
Tomas found that on ruby 2.0.0 the performance difference between fast_gettext 1.0.0 and 1.2.0 is much more significant.

PR #199, opened 2016-03-22
Fixes #14311 - Now hammer recognizes when to create defaults file correctly

ShimShtein:


tstrachota:
:+1:  thanks @ShimShtein !

PR #202, opened 2016-04-12
Fixes #14591 - Introduced log format setting

lzap:
By default the log format is wide, I prefer shorter format for development
purposes.


tstrachota:
Merged as f5a7c4e382ab5404e21c1a86f2786a487b3ed663 (I grouped the log related settings in the template config file together)
Thank you @lzap !

PR #203, opened 2016-04-14
Refs #8015 - allow param names to contain dashes

mbacovsky:
It seems more intuitive to use the param names in form they appear on
the command-line (organization_id vs. organization-id). This patch is
adding support for names with dashes.

How to test:
- make sure organization_id is not in defaults
  - `hammer defaults list`
- add organization-id
  - `hammer defaults add --param-name organization-id --provider foreman`
- call command requiring organization-id and check it worked
  - `hammer content-view list`

### DIFF
--- a/lib/hammer_cli/defaults.rb
+++ b/lib/hammer_cli/defaults.rb
@@ -24,7 +24,7 @@ module HammerCLI
 
     def delete_default_from_conf(param)
       conf_file = YAML.load_file(path)
-      conf_file[:defaults].delete(param)
+      conf_file[:defaults].delete_if { |k,| k.to_s.gsub('-','_') == param.to_s.gsub('-','_') }
       write_to_file conf_file
       conf_file
     end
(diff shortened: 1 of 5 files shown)


tstrachota:
Looks and works fine, @mbacovsky I only found one case when the behaviour seems weird:
```
hammer defaults add --param-name operatingsystem_id --param-value 83
hammer defaults add --param-name operatingsystem-id --param-value 4
```
```
hammer defaults list
-------------------|------
PARAMETER          | VALUE
-------------------|------
operatingsystem-id | 4    
operatingsystem_id | 83   
-------------------|------
```
And then the underscore version takes precedence. It's probably a minor thing but I'd expect the params to overwrite each other and be displayed only once in the list. What do you think?

mbacovsky:
I was aware of the problem but I preferred to leave it as transparent as possible. But as you say it I think you are right that replacing the value makes more sense. I updated the PR:

```text
$ hammer defaults list
----------------|---------------------
PARAMETER       | VALUE               
----------------|---------------------
organization-id | Provided by: Foreman
----------------|---------------------

$ hammer defaults add --param-name organization_id --provider foreman
Added organization_id default-option with value that will be generated from the server.

$ hammer defaults list
----------------|---------------------
PARAMETER       | VALUE               
----------------|---------------------
organization_id | Provided by: Foreman
----------------|---------------------
```

tstrachota:
Thanks @mbacovsky update works well now. I noticed there's the same issue with delete. 
```
hammer defaults add --param-name operatingsystem-id --param-value 4
hammer defaults delete --param-name operatingsystem_id
Couldn't find the requested param in /home/vagrant/.hammer/defaults.yml.
```
It's imho not that serious since users should know what they're deleting, but I'd expect the commands to be consistent.

mbacovsky:
@tstrachota updated, sorry I amended the changes by mistake :panda_face:

lib/hammer_cli/defaults.rb
```diff
@@ -35,20 +35,28 @@ def add_defaults_to_conf(default_options, provider)
...
-        defaults[:defaults][key] = value ? {:value => value,} : {:provider => provider}
+        defaults[:defaults].delete_if { |k,| k.to_s.gsub('-','_') == key.to_s.gsub('-','_') }
+        defaults[:defaults][key] = (value ? {:value => value,} : {:provider => provider})
       end
       write_to_file defaults
       defaults
     end
 
+    def defaults_set?(param)
+      defaults_settings.keys.any? { |k| k.to_s.gsub('-','_') == param.to_s.gsub('-','_') }
```
tstrachota: Could we get this one a bit more dry, please? It's similar pattern at 3 places.

tstrachota:
Works good now, @mbacovsky +1 to merge when my comment is addressed.

tstrachota:
:+1: thanks @mbacovsky :sunflower: :dromedary_camel:

PR #205, opened 2016-05-24
Fixes #14598 - Descriptions from Apipie will not contain unescaped HTML.

zjhuntin:
Currently Apipie only sends descriptions in escaped HTML. This causes the descriptions to look strange on escaped characters like ', & and others.

### DIFF
--- a/lib/hammer_cli/apipie/option_definition.rb
+++ b/lib/hammer_cli/apipie/option_definition.rb
@@ -11,6 +11,9 @@ module HammerCLI::Apipie
         self.referenced_resource = options.delete(:referenced_resource).to_s if options[:referenced_resource]
       end
       super
+      # Apipie currently sends descriptions as escaped HTML once this is changed this should be removed.
+      # See #15198 on Redmine.
+      @description = CGI::unescapeHTML(description)
     end
 
   end
--- a/test/unit/apipie/option_definition_test.rb
+++ b/test/unit/apipie/option_definition_test.rb
@@ -22,5 +22,14 @@ describe HammerCLI::Apipie::OptionDefinition do
     end
 
   end
+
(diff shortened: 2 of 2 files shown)


daviddavis:
Given that apipie is sending escaped html in its json, I think the fix needs to be in apipie and not hammer-cli.

daviddavis:
Ping @iNecas. Talking with @zjhuntin we were wondering—apipie is returning html entities (e.g. `	&#38;`) in the param descriptions within the json file. Is that intentional? Is anything expecting html in the json?

Relevant code links:

https://git.io/vr6IX
https://git.io/vr6IM

iNecas:
Yes, the problem right now in apipie is we use the to_json method in both html and json version https://github.com/Apipie/apipie-rails/blob/296c097f9acb951149d3a67db69b84eb216bdd7e/app/controllers/apipie/apipies_controller.rb#L34. To change this, it would mean an untrivial amount of work to customize the generation based on the purpose (html vs. pure json).

zjhuntin:
This fix should be removed once, https://git.io/vr17V has been completed in Apipie. There as been a issue made for this in Redmine as well, http://projects.theforeman.org/issues/15198.

tstrachota:
There's the same problem with command descriptions that come form apipie, @zjhuntin. Can you please add similar sanitization here https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/apipie/command.rb#L16 ?

zjhuntin:
@tstrachota I think that the necessary changes have been made.

tstrachota:
Cool, @zjhuntin May I ask you for a simple test for the descriptions too, please?

zjhuntin:
@tstrachota So I am actually unsure of what that description normally should be. If it is always supposed to be the short description provided by Apipie then it does not need to be escaped. Apipie only converts the longer descriptions into escaped HTML before sending it over. 

From what I see I think it would only get the short descriptions, is that incorrect?

tstrachota:
@zjhuntin I see, I didn't know short descriptions are not affected. Ok then, merging. Thank you!

PR #207, opened 2016-06-03
Fixes #13832 - richer option validations

tstrachota:
Option validator gets:
- support for testing option values
- better messages for single option checks
- support for testing is one of options is present

### DIFF
--- a/lib/hammer_cli/validator.rb
+++ b/lib/hammer_cli/validator.rb
@@ -89,6 +89,39 @@ module HammerCLI
     end
 
 
+    class OneOfConstraint < BaseConstraint
+      def initialize(options, to_check)
+        raise 'Set at least one expected option' if to_check.empty?
+        super(options, to_check)
+      end
+
+      def rejected
+        raise NotImplementedError, '#rejected is unsupported for OneOfConstraint'
+      end
+
+      def required_msg
+        case count_present_options
+        when 0
+          _("One of options %s is required")
+        when 1
+          ''
+        else
+          _("Only one of options %s can be set")
+        end
+      end
+
+      def exist?
+        return count_present_options == 1
+      end
+
+      protected
+      def count_present_options
+        @to_check.select do |opt|
+          option_passed?(opt)
+        end.count
+      end
+    end
+
     def initialize(options)
       @options = options
     end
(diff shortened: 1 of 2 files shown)


lib/hammer_cli/validator.rb
```diff
@@ -89,6 +100,39 @@ def exist?
     end
 
 
+    class OneOfConstraint < BaseConstraint
+      def initialize(options, to_check)
+        raise 'Set at least one expected option' if to_check.empty?
```
adamruzicka: Translations are not needed here?

tstrachota: It's a runtime exception for developers so I didn't translate it.

lib/hammer_cli/validator.rb
```diff
@@ -98,13 +142,17 @@ def all(*to_check)
     end
 
     def option(to_check)
-      all(to_check)
+      OneOptionConstraint.new(@options, to_check)
```
adamruzicka: This might eventually cause failing tests in various hammer plugins (like here https://github.com/theforeman/hammer-cli-foreman/pull/233/files#diff-dbd90652e9404f49d32e20e0c7f4a976L31 ). I checked `hammer-cli-{foreman,foreman-tasks,katello}` and `hammer_cli_{foreman_remote_execution,foreman_docker}`, those aren't affected by this change

adamruzicka:
Left few small comments inline, both getting `OneOfConstraint` and getting option values in validations works great except for retrieving option values in tests, created [issue](http://projects.theforeman.org/issues/15298) for that

lib/hammer_cli/validator.rb
tstrachota: Thanks for checking it! It's breaking change but since it changes only the messages and it imho changes them for better, I'd keep it in.

tstrachota:
@adamruzicka thanks, updated. I looked at the mentioned issue and it turned out to be an apidoc problem. We use old export in the tests. I changed the issue title and the description.

lib/hammer_cli/validator.rb
adamruzicka: Sure, nothing against that. Just thought it might be a good idea to have a note about this kept somewhere

adamruzicka:
@tstrachota Thanks for looking into that issue. Code looks good and works well, :+1:

tstrachota:
Thanks @adamruzicka, merging

tstrachota:
Merged as def5d036c660682e39ffd8d3dd322bfca7769630 and 5456fe2a878d573eb3fcb0f73ad69c84483aec12

PR #208, opened 2016-06-06
Fixes #15257 - let print adapters decide whether to paginate

tstrachota:


tstrachota:
Small update: I removed the default pagination setting from the config file template.

akofink:
Looks great; I like the design. The tests pass locally and seem to cover the functionality implemented.

ACK from me, though I'm not in the Foreman org

lib/hammer_cli/output/adapter/abstract.rb
```diff
@@ -11,6 +11,10 @@ def initialize(context={}, formatters={})
       @formatters = HammerCLI::Output::Formatters::FormatterLibrary.new(filter_formatters(formatters))
     end
 
+    def paginate_by_default?
+      true
+    end
+
```
mbacovsky: Nit: I'd prefer to refactor to
```ruby 
def paginate_by_default?
    !!@paginate_by_default
end
```
and set the attribute in initializer, but it is minor thing

tstrachota:
@mbacovsky ok, updated

mbacovsky:
:+1: Thanks @tstrachota! :poodle:

PR #209, opened 2016-06-15
i18n - fix extraction of variables in strings

domcleal:


mbacovsky:
:+1: Thanks @domcleal!

PR #210, opened 2016-06-22
refs #7453 - explicitly list man page in gemspec files

domcleal:
Ensures that the man page is built and included in the gem if `gem
build` rather than `rake build` is used.


mbacovsky:
Good idea, thanks @domcleal!
:+1:

PR #206, opened 2016-06-01
fixes #14436 - added exception handling while parsing configuration file.

Akasurde:
Notifies user about syntax error in given configuration file
provided.

Fixes: http://projects.theforeman.org/issues/14436

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

PR #211, opened 2016-08-18
Fixes #14436 - exception handling for parsing configuration - take 2

tstrachota:
Since the [original PR](https://github.com/theforeman/hammer-cli/pull/206) has been silent for a while I'm opening a new one with with additional commit that should address the comments.

### DIFF
--- a/lib/hammer_cli/settings.rb
+++ b/lib/hammer_cli/settings.rb
@@ -40,8 +40,8 @@ module HammerCLI
             load(config)
             path_history << file_path
           end
-        rescue Exception
-          puts _("Couldn't load configuration file %s, please check the YAML file for syntax.") % file_path
+        rescue Exception => e
+          warn _("Warning: Couldn't load configuration file %{path}: %{message}") % { path: file_path, message: e.message }
         end
       end
     end
--- a/lib/hammer_cli/settings.rb
+++ b/lib/hammer_cli/settings.rb
@@ -34,10 +34,14 @@ module HammerCLI
 
     def self.load_from_file(file_path)
       if File.file? file_path
-        config = YAML::load(File.open(file_path))
-        if config
-          load(config)
-          path_history << file_path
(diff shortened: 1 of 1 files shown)


lib/hammer_cli/settings.rb
```diff
@@ -40,8 +40,8 @@ def self.load_from_file(file_path)
             load(config)
             path_history << file_path
           end
-        rescue Exception
-          puts _("Couldn't load configuration file %s, please check the YAML file for syntax.") % file_path
+        rescue Exception => e
+          puts _("Couldn't load configuration file %{path}: %{message}") % { path: file_path, message: e.message }
```
mbacovsky: After some testing I'd prefer to use warning `warn _("Warning: Couldn't......`.
It feels better visible and more in line with other similar output we do. What do you think?

tstrachota: I agree. In fact I overlooked there's puts instead of warn.

tstrachota:
@mbacovsky updated

lib/hammer_cli/settings.rb
```diff
@@ -34,10 +34,14 @@ def self.load_from_paths(files)
...
-          load(config)
-          path_history << file_path
+        begin
+          config = YAML::load(File.open(file_path))
+          if config
+            load(config)
+            path_history << file_path
+          end
+        rescue Exception => e
+          warn _("Couldn't load configuration file %{path}: %{message}") % { path: file_path, message: e.message }
```
mbacovsky: Almost perfect, could you please prefix the message  with "Warning: " as we do elsewhere.

tstrachota:
...a šmidra vydra, @mbacovsky

mbacovsky:
Significantly better, Bobe! :panda_face: 
:+1:
Will be merged without squashing after checks finish.

PR #212, opened 2016-08-18
Fixes #16161 - option deprecation indication in help

tstrachota:


iNecas:
When testing with https://github.com/theforeman/hammer-cli-foreman/pull/256, I would expect also to see what to use instead (the string of :deprecated, warning)

tstrachota:
I updated the code to print the message when the whole option (with all switch aliases) is deprecated.

@mbacovsky is the partial deprecation of certain switches used anywhere? It feels quite unnatural.

mbacovsky:
@tstrachota, the partial deprecation is used when we wanted to rename certain options and wanted to keep compatibility with previous version. Alias with old name that was deprecated solved that. It was used downstream only. But still can be useful feature of the option builder.

It makes sense to show deprecated only for fully deprecated options but theoretically `option ['--opt'], :deprecated => "message"` and `option ['--opt'], :deprecated => {'--opt' => "message"}` are semantically same. You could print something like `--opt, --old-opt description (--old-opt is deprecated)` for options with :deprecated being hash

tstrachota:
@mbacovsky makes sense, updated.

mbacovsky:
Works great now, thanks @tstrachota! :green_apple: 
:+1:

PR #217, opened 2016-09-21
Fixes #16631 - Use defaults in validators

mbacovsky:


tstrachota:
:+1: works fine. Thanks for the tests @mbacovsky

PR #220, opened 2016-10-05
Fixes #16811 - Allow multiple opt validation blocks

daviddavis:
I have a module that defines some option validation and it can be called multiple times. In order to do this, I need to be able to set multiple validation blocks. I tried to write something that could be called from within 'validation_options' but the code there is executed within HammerCLI::Validator so I don't think that'll work.

## Todo
- [x] Add tests

### DIFF
--- a/lib/hammer_cli/abstract.rb
+++ b/lib/hammer_cli/abstract.rb
@@ -11,7 +11,7 @@ module HammerCLI
     include HammerCLI::Subcommand
 
     class << self
-      attr_accessor :validation_block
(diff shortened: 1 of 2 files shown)


daviddavis:
Related PR: https://github.com/Katello/hammer-cli-katello/pull/459

tstrachota:
Nice addition @daviddavis !

daviddavis:
@tstrachota updated with a test.

tstrachota:
Merged, thanks @daviddavis

PR #221, opened 2016-10-06
Fixes #16822 - Check for nil default values

akofink:
https://github.com/theforeman/hammer-cli/pull/217 introduces failing tests to hammer-cli-foreman and hammer-cli-katello. This nil check fixes both projects.

### DIFF
--- a/lib/hammer_cli/validator.rb
+++ b/lib/hammer_cli/validator.rb
@@ -45,7 +45,10 @@ module HammerCLI
       def get_option_value(name)
         opt = get_option(name)
         value = opt.get
-        value = opt.command.send(:context)[:defaults].get_defaults(name.to_s) if value.nil?
+        if value.nil?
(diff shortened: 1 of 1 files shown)


lib/hammer_cli/validator.rb
```diff
@@ -45,7 +45,10 @@ def get_option(name)
       def get_option_value(name)
         opt = get_option(name)
         value = opt.get
-        value = opt.command.send(:context)[:defaults].get_defaults(name.to_s) if value.nil?
+        if value.nil?
+          defaults = opt.command.send(:context)[:defaults]
+          defaults.get_defaults(name.to_s) if defaults
```
mbacovsky: `value = defaults.get_defaults(name.to_s) if defaults`

mbacovsky:
:+1: Thanks for the fix, @akofink!

PR #223, opened 2016-10-17
Fixes #16973 - Handle method refactor in clamp 1.10

daviddavis:
More info: https://github.com/mdub/clamp/commit/3d3788a799b88ab0433e4a7ff8fd131144dcfec5

### DIFF
--- a/hammer_cli.gemspec
+++ b/hammer_cli.gemspec
@@ -23,7 +23,7 @@ EOF
   s.require_paths = ["lib"]
   s.executables = ['hammer']
 
-  s.add_dependency 'clamp', '>= 1.0.0'
+  s.add_dependency 'clamp', '~> 1.0.0'
   s.add_dependency 'logging'
   s.add_dependency 'awesome_print'
   s.add_dependency 'table_print'


lib/hammer_cli/abstract.rb
```diff
@@ -137,7 +137,7 @@ def self.build_options(builder_params={})
 
         declared_options << option
         block ||= option.default_conversion_block
-        define_accessors_for(option, &block)
+        declare_attribute(option, &block)
```
mbacovsky: @daviddavis it won't work with clamp < 1.1.0.

mbacovsky:
Renaming the methods in Hammer would lock us to clamp 1.1.0. We need to support < 1.1.0 too.
I also checked that 1.1.0 also drops ruby 2.0.0 support.So I'd prefer to lock clamp to < 1.1.0 for now.

daviddavis:
@mbacovsky updated

mbacovsky:
:+1: Works well now, thanks, @daviddavis !

mbacovsky:
@daviddavis could you please change the commit message to match the change?

mbacovsky:
@daviddavis thanks! Merging...

PR #224, opened 2016-10-19
refs #16973 - permit clamp 1.1 and higher, compatibility restored

domcleal:


daviddavis:
LGTM

hammer_cli.gemspec
```diff
@@ -23,7 +23,7 @@ EOF
   s.require_paths = ["lib"]
   s.executables = ['hammer']
 
-  s.add_dependency 'clamp', '~> 1.0.0'
+  s.add_dependency 'clamp', '~> 1.0'
```
mbacovsky: Is there a nice way we can exclude `1.1.0` from the range? It introduced some refactoring causing failures and it was taken back again in `1.1.1`

mbacovsky:
:+1:  Let's keep it as you suggest. Thanks @domcleal!

PR #225, opened 2016-10-25
Fixes #17021 - properly detect booleans

ares:


ares:
@tstrachota or @mbacovsky would you mind taking a look? should not take more than a minute to review, the issue was that `param.expected_type` could be `:boolean` (symbol vs string)

lib/hammer_cli/apipie/option_builder.rb
```diff
@@ -66,7 +66,7 @@ def option_opts(param)
       opts[:required] = true if (param.required? and require_options?)
       if param.expected_type == :array || param.validator =~ /Array/i
```
domcleal: Use the same style here, either convert to a string like with boolean or use a symbol comparison below.

ares:
Thanks @domcleal, updated both array and number types.

domcleal:
Thanks, I'm happy with the change but would only ask for the tests to be updated too. It looks like this does actually have some test coverage but it relies on the _validator_ being detected rather than the expected_type of the field.

Check test/unit/fixtures/apipie/documented.json and compare it to a more recent apipie-rails output from Foreman and the validator field appears to be different (`Must be Boolean` is now `Must be 'true' or 'false' or '1' or '0'`). It's probably best not to rely on the validator field, but to update the validators in the fixtures to match a more current version which should ensure the test checks the same thing.

ares:
@tstrachota do you have some app in which you generated that `documented.json` fixture? Could you regenerate it with newer apipie?

tstrachota:
@ares no, it was created manually. I usually copy snippets from Foreman's generated apipie cache.

ares:
@domcleal I'm afraid I don't entirely understand what you mean by ` It's probably best not to rely on the validator field`. I updated the fixture. If you want me to add something more, please give me some hints or code pointers. I'm not too familiar with hammer, so I'd like to avoid bigger changes though.

domcleal:
That was all! Thanks @ares.

PR #226, opened 2016-11-02
Fixes #17180 - Double quotes in list type params

mbacovsky:
`--parameter='a,b,"c,d"'` gets parsed as `['a', 'b', 'c,d']`

### DIFF
--- a/lib/hammer_cli/options/normalizers.rb
+++ b/lib/hammer_cli/options/normalizers.rb
@@ -1,4 +1,5 @@
 require 'json'
+require 'csv'
 
 module HammerCLI
   module Options
@@ -82,11 +83,11 @@ module HammerCLI
       class List < AbstractNormalizer
 
         def description
-          _("Comma separated list of values.")
+          _("Comma separated list of values. Values containing comma should be double quoted")
         end
 
         def format(val)
-          val.is_a?(String) ? val.split(",") : []
+          val.is_a?(String) ? CSV.parse(val).flatten(1) : []
         end
       end
 
--- a/lib/hammer_cli/options/normalizers.rb
+++ b/lib/hammer_cli/options/normalizers.rb
@@ -79,6 +79,13 @@ module HammerCLI
         end
       end
 
+      CSV_ERROR_MESSAGES = [
+        N_('Missing or stray quote in line %s.'),
+        N_('Unquoted fields do not allow \r or \n (line %s).'),
+        N_('Illegal quoting in line %s.'),
+        N_('Unclosed quoted field on line %s.'),
+        N_('Field size exceeded on line %s.')
+      ]
 
       class List < AbstractNormalizer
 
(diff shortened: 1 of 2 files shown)


tstrachota:
It parses the values well. I found one small issue with reporting malformed csv:
```
[~|vm 2.3]hammer domain create --name test.com --organization-ids='1,"3,4""'                                                                
Could not create the domain:
  Error: Unclosed quoted field on line 1.
  
[~|vm 2.3]hammer domain create --name test.com --organization-ids='1,3,4""'
Could not create the domain:
  Error: Illegal quoting in line 1.
```

It would be good to catch such exceptions and re-raise `ArgumentError` with appropriate message. That will give us error like:
```
Could not create the domain:
  Error: option '--organization-ids': illegal quoting.
  
  See: 'hammer domain create --help'
```

mbacovsky:
Good idea @tstrachota, updated.
```
$ hammer domain create --name test.com --organization-ids='1,"3,4""'
Could not create the domain:
  Error: option '--organization-ids': Unclosed quoted field on line 1.
  
  See: 'hammer domain create --help
$ hammer domain create --name test.com --organization-ids='1,3,4""'
Could not create the domain:
  Error: option '--organization-ids': Illegal quoting in line 1.
  
  See: 'hammer domain create --help'
```

lib/hammer_cli/options/normalizers.rb
```diff
@@ -82,11 +83,13 @@ def strip_chars(string, chars)
       class List < AbstractNormalizer
 
         def description
-          _("Comma separated list of values.")
+          _("Comma separated list of values. Values containing comma should be double quoted")
         end
 
         def format(val)
-          val.is_a?(String) ? val.split(",") : []
+          val.is_a?(String) ? CSV.parse(val).flatten(1) : []
```
tstrachota: Btw I found out you can do `CSV.parse_line(val)` and get rid of `flatten`.

tstrachota:
Great, thanks for the update @mbacovsky 

I found out the messages aren't translated. Do you think we should detect the messages and print translated form (without the line number which is always going to be 1)?

It's only 5 strings (see https://github.com/ruby/ruby/blob/trunk/lib/csv.rb):
* "Missing or stray quote in line #{lineno + 1}"
* "Unquoted fields do not allow " + "\\r or \\n (line #{lineno + 1})."
* "Illegal quoting in line #{lineno + 1}."
* "Unclosed quoted field on line #{lineno + 1}."
* "Field size exceeded on line #{lineno + 1}."

mbacovsky:
@tstrachota, rebased, `parse_line` used. 

I've tried to approach the translations of the errors. This is probably most minimal version I came up with. I'm not sure if the module is best place where to put the strings to be noted by gettext. Any better idea?

tstrachota:
Having the strings in the module shouldn't be any problem.

An alternative approach that came to my mind is having the translations stored in a hash of `regex => translation`. This way would give us more flexibility in what to put into translated messages. I don't think it's necessary to copy the information about line, because it will be always 1 in the case of our parameter input.

```ruby
CSV_ERROR_MESSAGES = {
  /Missing or stray quote/ => _('Missing or stray quote.'),
  /Unquoted fields/ => _('Unquoted fields do not allow \r or \n .')
  #...
}
# ... then search for matching key and use value for the error message.
```

mbacovsky:
@tstrachota, I find your idea more user friendly, please check the update.

tstrachota:
Thanks @mbacovsky

PR #227, opened 2016-11-24
Fixes #8016 - api connection moved to context

tstrachota:
- connection moved to context
- connection is now instance, not class

TODO: 
- [x] fix tests

### DIFF
--- a/lib/hammer_cli/context.rb
+++ b/lib/hammer_cli/context.rb
@@ -3,9 +3,10 @@ require 'hammer_cli/defaults'
 module HammerCLI
 
   def self.context
-    {
+    @context ||= {
       :defaults => HammerCLI.defaults,
-      :is_tty? => HammerCLI.tty?
+      :is_tty? => HammerCLI.tty?,
+      :api_connection => HammerCLI::Connection.new(Logging.logger['Connection'])
     }
   end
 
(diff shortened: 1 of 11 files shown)


lib/hammer_cli/context.rb
```diff
@@ -3,9 +3,10 @@
 module HammerCLI
 
   def self.context
-    {
+    @context ||= {
       :defaults => HammerCLI.defaults,
-      :is_tty? => HammerCLI.tty?
+      :is_tty? => HammerCLI.tty?,
+      :api_connection => HammerCLI::Connection.new(Logging.logger['Connection'])
```
mbacovsky: Would it make sense to move this to command initializer and add just the attributes missing? This way you could provide only part of the context to the command and the rest would be replaced with the defaults. The context would also be accessible with `@context`?

mbacovsky:
The context is not blocker for getting this merged. It just occurred to me that while Clamp allows to provide the context as command parameter. We would not be able to add the api_connection when it is missing there.

I'd like to know if you think it should be fixed in this PR.  Otherwise I'm going to merge it.

lib/hammer_cli/context.rb
tstrachota: I'm not sure I understand. The context is passed to commands' constructors and is accessible as `self.context` or `@context`.
I don't think setting a default value for the api connection makes sense in this case, because we need one connection for the whole hammer.

mbacovsky: I see, sorry for the confusion.

mbacovsky:
:+1: Thanks @tstrachota!

PR #229, opened 2016-12-20
Fixes #17740 - replacement for table_print

tstrachota:
Replaces table print with a custom solution due to performance reasons.  See the redmine issue http://projects.theforeman.org/issues/17740 for details about the current performance measurements. This change should make tabular output as fast as other adapters.

The patch should be 100% compatible with the previous version of table adapter in terms of produced output. Please test it thoroughly to catch possible deviations from the original formatting.


tstrachota:
@akofink may I ask you for review?

lib/hammer_cli/output/adapter/table.rb
```diff
@@ -68,73 +58,68 @@ def print_collection(all_fields, collection)
...
+        value += ' ' if real_length < (width - 3)
+      end
+      value
     end
 
-    private
+    def format_values(fields, collection)
+      collection.collect do |d|
+        row = {}
+        fields.each do |f|
```
ares: I think this is equivalent
```ruby
fields.inject({}) do |row, f|
  formatter = WrapperFormatter.new(@formatters.formatter_for_type(f.class), f.parameters)
  row[f.label] = formatter.format(data_for_field(f, d) || "").to_s
end
```

lib/hammer_cli/output/adapter/table.rb
```diff
@@ -68,73 +58,68 @@ def print_collection(all_fields, collection)
...
 
-    def label_for(field)
-      width = width_for(field)
-      if width
-        "%-#{width}s" % field.label.to_s
-      else
-        field.label.to_s
+    def calculate_widths(fields, collection)
+      widths = {}
+      fields.map do |f|
```
ares: I think this can be onliner, but feel free to ignore if you find it unreadable
```ruby
Hash[fields.map { |f| [f.label, calculate_column_width(f, collection) }]
```

akofink:
It's so much faster (try `hammer erratum list`), and it functions flawlessly with hammer-cli-katello. ACK from me. Thanks @tstrachota!

tstrachota:
@ares updated to use cooler syntax when working with hashes ;) If you ack I'm going to merge it.

ares:
:+1:

tstrachota:
Merged, thanks for the review @akofink @ares

PR #230, opened 2017-01-12
Fixes #12401 - Add support for SSL options

rwf14f:
Add SSL options to allow:
- verification of server certificate (fixes http://projects.theforeman.org/issues/12400)
- client authentication (fixes http://projects.theforeman.org/issues/12401)


ares:
ok to test

mmoll:
For hammer I think/hope it's still
[test]

bin/hammer
```diff
@@ -31,6 +31,11 @@ class PreParser < Clamp::Command
   option ["--output"], "ADAPTER", "Set output format"
   option ["--csv-separator"], "SEPARATOR", "Character to separate the values"
   option ["--autocomplete"], "LINE", "Get list of possible endings"
+  option ["--verify_ssl"], "VERIFY_SSL", "Configure SSL verification of remote system"
+  option ["--ssl_ca_file"], "CA_FILE", _("configure the file containing the CA certificates")
+  option ["--ssl_ca_path"], "CA_PATH", _("configure the directory containing the CA certificates")
+  option ["--ssl_client_cert"], "CERT_FILE", _("configure the client's public certificate")
+  option ["--ssl_client_key"], "KEY_FILE", _("configure the client's private key")
```
tstrachota: Can you please replace underscores with hyphens in the options? Hammer doesn't use underscores anywhere else in parameters.
`--ssl_client_key` -> `--ssl-client-key`

lib/hammer_cli/ssloptions.rb
```diff
@@ -0,0 +1,27 @@
+module HammerCLI
+  def self.get_ssl_options
+    settings = HammerCLI::Settings
+    logger = Logging.logger['SSLoptions']
```
tstrachota: Do you think you could move `logger` and `settings` to the function constructor and make it injectable dependencies? It would make testing easier and more straightforward.

lib/hammer_cli/ssloptions.rb
```diff
@@ -0,0 +1,27 @@
...
+      ssloptval = settings.get(:_params, sslopt) || settings.get(:ssl, sslopt)
+      ssl_options[sslopt] = ssloptval if ssloptval
+    end
+    ssl_client_cert = settings.get(:_params, :ssl_client_cert) || settings.get(:ssl, :ssl_client_cert)
+    ssl_client_key = settings.get(:_params, :ssl_client_key) || settings.get(:ssl, :ssl_client_key)
+    if ssl_client_cert && ssl_client_key
+      ssl_options[:ssl_client_cert] = OpenSSL::X509::Certificate.new(File.read(ssl_client_cert))
+      ssl_options[:ssl_client_key] = OpenSSL::PKey::RSA.new(File.read(ssl_client_key))
+    elsif ssl_client_cert
+      logger.warn("SSL client certificate is set but the key is not, SSL client authentication disabled")
```
tstrachota: I think this message is important enough to display it to users as a standard warning:
warn _("SSL client...")

lib/hammer_cli/ssloptions.rb
```diff
@@ -0,0 +1,27 @@
...
+    elsif ssl_client_key
+      logger.warn("SSL client key is set but the certificate is not, SSL client authentication disabled")
```
tstrachota: ditto

lib/hammer_cli/ssloptions.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+    verify_ssl = settings.get(:_params, :verify_ssl) || settings.get(:ssl, :verify_ssl)
+    ssl_options[:verify_ssl] = verify_ssl unless verify_ssl.nil?
+    # enable ssl verification if verify_ssl is not configured and either CA file or path are present
+    ssl_options[:verify_ssl] = 1 if ssl_options[:verify_ssl].nil? && (ssl_options[:ssl_ca_file] || ssl_options[:ssl_ca_path])
+    logger.debug("ssl options: #{ApipieBindings::Utils::inspect_data(ssl_options)}")
```
tstrachota: Useful, thanks for logging that!
Just minor nitpick: We start all log messages with capital letters.

lib/hammer_cli/ssloptions.rb
```diff
@@ -0,0 +1,27 @@
+module HammerCLI
+  def self.get_ssl_options
```
tstrachota: This method is quite complex. Could you add a unittest? Either a standalone one or you can test it as part of the `ApiConnection` class in `test/unit/apipie/api_connection_test.rb `.

bin/hammer
tstrachota: Btw there's no need to translate the option descriptions here. They exist only for documentation purposes. We use options from main command to print help.

lib/hammer_cli/ssloptions.rb
```diff
@@ -0,0 +1,27 @@
...
+    logger = Logging.logger['SSLoptions']
+    ssl_options = {}
+    for sslopt in [:ssl_ca_file, :ssl_ca_path] do
+      ssloptval = settings.get(:_params, sslopt) || settings.get(:ssl, sslopt)
+      ssl_options[sslopt] = ssloptval if ssloptval
+    end
+    ssl_client_cert = settings.get(:_params, :ssl_client_cert) || settings.get(:ssl, :ssl_client_cert)
+    ssl_client_key = settings.get(:_params, :ssl_client_key) || settings.get(:ssl, :ssl_client_key)
+    if ssl_client_cert && ssl_client_key
+      ssl_options[:ssl_client_cert] = OpenSSL::X509::Certificate.new(File.read(ssl_client_cert))
```
tstrachota: Please catch exception for file read errors to provide meaningful message. Non-existent file currently makes hammer to fail:
```
hammer --ssl_client_key ~/unknown/key --ssl_client_cert ~/unknown/cert  user list
Error: No such file or directory @ rb_sysopen - /root/unknown/cert
Warning: An error occured while loading module hammer_cli_foreman
Warning: An error occured while loading module hammer_cli_foreman_tasks
Error: No such file or directory @ rb_sysopen - /root/unknown/cert
Warning: An error occured while loading module hammer_cli_katello
Error: No such sub-command 'user'

See: 'hammer --help'
```

tstrachota:
@rwf14f I finally managed to do some deeper testing. Apart from what I've already wrote in the review I found some more issues:
- `--verify-ssl=1` caused failures in rest-client gem when passed from command line, because tha value was string
- new settings should be reflected in the default config file
- apipie-bindings is hiding ssl verification failures during cache retrieval (https://github.com/Apipie/apipie-bindings/pull/57)

Since this is really cool features and I'd love to get them into 0.10.0 release I fixed the issues I found and made some refactoring for better testability. You can find my patches in: https://github.com/rwf14f/hammer-cli/pull/1 Please look at them and merge what you find useful. Tests are still missing.

Can you please fix your commit messages to standard format "Fixes #12401 - ..."?

I'd say that `ssl_client_cert` and `ssl_client_key` should probably be set from an authenticator. Otherwise they will be used also for unauthenticated calls like fetching cache. @mbacovsky what do you think? Is it a big deal?

rwf14f:
@tstrachota Thanks for implementing the changes, I currently don't have much time to look at this. I've merged them and updated the commit messages and the title of the pull request.

In theory I'd agree with you that the correct place to configure the client certificate/key authentication is an authenticator. I don't see how this could work with current implementation of the authenticators though. The problem is that Foreman is running behind Apache which is configured to accept only authenticated connections (by providing a client certificate/key). This includes connections for which Foreman doesn't require authentication itself, e.g. the initial call to /apidoc/... of the foreman client.

tstrachota:
Ah, you're right, @rwf14f . We'd need some mechanism to detect whether the authenticator needs to cover apidoc requests too. I'm fine with the current implementation then.

Since I've added some code, I'd like a third person to review and ack before we merge.

mbacovsky:
@tstrachota, @rwf14f I can do the "unbiased" review ;)

tstrachota:
BTW I created a PR into hammer-cli-foreman with updated exception handling for apipie doc-loading errors https://github.com/theforeman/hammer-cli-foreman/pull/291

mbacovsky:
@rwf14f, The code looks good to merge, thanks for your effort!

@rwf14f, @tstrachota  
I found one marginal issue that I'd like to discuss:
I configured foreman to auth user by client cert:
 - in Settings/Authentication  > Authorize Login delegation Yes
 - in Settings/Authentication  > Authorize Login delegation API  Yes
 - in `/etc/httpd/conf.d/05-foreman-ssl.conf` add `SSLUserName SSL_CLIENT_S_DN_CN`
 - restart httpd
 - create user test
 - generate the client cert `puppet cert --generate test`

Then 
```
hammer --verify-ssl true --ssl-ca-file ca.pem --ssl-client-cert test.pem --ssl-client-key test.key host list
```
asked me for username and password. The user taken from cert CN takes precedence and my input is ignored. It could be workarounded by adding fake username and password to the hammer config, but I'd prefer not to be asked for username and password if cert is provided. Makes sense?

mbacovsky:
[test]

mbacovsky:
@tstrachota, I like it. My only concern is that it does not allow a case when a user wants to use the client cert just for SSL and still auth with his credentials. I'm not sure if this is valid and meaningful usecase though. If it is I'd suggest to add `auth_with_client_cert` option/config entry and handle the credentials accordingly. Any opinion @rwf14f ?

rwf14f:
I agree.  Certificate and password authentication are not mutually exclusive. I can't think of a good reason why a user would want to do this though either (maybe there's a security policy requiring it). I think it would be good to allow it. My first thought was to use the VoidAuth only if no user name / password are specified, but that would prevent the user from being able to specify them interactively. An additional option as you suggested is better.

mbacovsky:
@rwf14f thanks for your input. Could you please add the additional option (with whatever suitable name) to the PR (hammer bin, main, and sample config)?

@tstrachota could you please update your hammer-cli-foreman PR accordingly when the change is ready?

I've created new related Issue to add documentation for this feature (http://projects.theforeman.org/issues/18930).

rwf14f:
I've added the option and had a go at implementing it to test it: #https://github.com/tstrachota/hammer-cli-foreman/pull/2/commits/9fa1cb0916b2c542661919936f458f90b6afe90d .
@tstrachota I've sent you a pull request if you want to use my implementation: https://github.com/tstrachota/hammer-cli-foreman/pull/2

tstrachota:
Cool @rwf14f Thank you! :1st_place_medal: 
This is very nice improvement that is worth showing off to the world. Would you be willing to show it at the next Foreman demo when we merge it?

rwf14f:
@tstrachota thanks for the offer, but I pass.
@mbacovsky Can you test the latest changes to make sure we haven't overlooked anything ?

mbacovsky:
[test]

mbacovsky:
@rwf14f, thanks for the update. I seems the option name I suggested can be confusing (didn't occur to me before). By `auth-with-client-cert` I meant 'Authenticate with client cert' while it seems you understood it as 'Basic auth  with client cert`.  

My intention was to use basic auth with client cert as a default and have a switch to force basic auth off. Your take  (client cert authentication by default and turning basic on by switch) might make better sense, but then I'd prefer different name. Or is it confusing just for me? @tstrachota?

tstrachota:
I definitely vote for having the base auth off by default when using ssl certificates. How about renaming the option (and setting) to `--ssl-with-base-auth` to make it more obvious?

tstrachota:
@mbacovsky I'm changing the name in: https://github.com/rwf14f/hammer-cli/pull/2
If you have nothing against, @rwf14f, please merge. The foreman's counter PR has been modified accordingly. I'm open to other names if you don't like it.

rwf14f:
No objections from me.  If @mbacovsky has no objections then I'll merge.

mbacovsky:
@rwf14f, merge it in please and I'll do final review/testing

mbacovsky:
[test]

mbacovsky:
:+1: Thanks @rwf14f and @tstrachota for your patience and effort. This feature passed my testing and can be merged. :cake: :wine_glass: :beers:

PR #232, opened 2017-02-09
Fixes #18446 - Hide label fields when the value is empty or nil

tstrachota:


lib/hammer_cli/output/fields.rb
```diff
@@ -85,7 +85,7 @@ class Label < ContainerField
     def display?(value)
       return true if not hide_blank?
 
-      fields.any? do |f|
+      !(value.nil? || value.empty?) && fields.any? do |f|
```
ares: Isn't this equivalent to `value.present?`

ares: oh, hammer probably does not depend on active support gem, sorry, dismiss :-)

shlomizadok:
LGTM :+1:

tstrachota:
Thanks guys, merging

PR #233, opened 2017-03-01
Fixes #17920 - defaults match on option switch name

tstrachota:
Method #get_default was changed to accept switch (`--options`) instead of attribute reader name. The internal format of how the defaults are saved remain the same.


test/unit/abstract_test.rb
```diff
@@ -351,6 +351,40 @@ def self.create_option_builder
 
   end
 
+  describe 'defaults' do
+    class TestDefaultsCmd < HammerCLI::AbstractCommand
+      option '--test', 'TEST', '', :attribute_name => :different_attr_name
+      option ['--test-multi1', '--test-multi2'], 'TEST_MULTI', '', :attribute_name => :multiple_switches_option
+
+      def options
+        super
```
tstrachota: The fact that I had to use this hack makes me think I should have extracted the code into a class to make it testable. I'll probably refactor that.

ares:
I didn't do the full code review but fixes the issue for me, :+1:

mbacovsky:
I'm half way through the review, I'll finish it today.

tstrachota:
@mbacovsky even if such options existed the attribute reader isn't visible to users. They see only option flags. So if I'm not wrong it can, at the worst case, confuse only developers. That's something I can live with.

tstrachota:
And yes, please wait for the refactoring. I'll do a quick POC that I'll send you to ack/nack.

tstrachota:
@mbacovsky I added a commit with the refactoring. I designed it with http://projects.theforeman.org/issues/17252 in mind. I'd like to have one more option source that replaces nil placeholders with real nil values.
Let me know what you think.

lib/hammer_cli/options/option_collector.rb
```diff
@@ -0,0 +1,22 @@
...
+      attr_accessor :option_sources
+
+      def initialize(recognised_options, option_sources)
+        @recognised_options = recognised_options
+        @option_sources = option_sources
+      end
+
+      def all_options
+        @all_options ||= @option_sources.inject({}) do |all_options, source|
+          source.get_options(@recognised_options, all_options)
```
mbacovsky: I've seen the `result` is evaluated in `SavedDefaults` option source but could each option source just return its values and get them merged together in order in which the sources are stored in the collector?

tstrachota: I want option sources to have access to the full option hash, because there can be situations when you need other options' values. For example whole name -> id resolution could be refactored into an option source.

mbacovsky: Makes sense. Thanks

mbacovsky:
:+1: thanks @tstrachota, merging :icecream:

PR #234, opened 2017-03-21
Use HTTPS for downloading the apt key

faxm0dem:


mmoll:
OK for me, but IMHO this also would need a `apt-get -y install ca-certificates` at least on Debian.

tstrachota:
Thanks @faxm0dem and @mmoll

PR #235, opened 2017-03-28
Fixes #19033 - Enable SSL verification by default

tstrachota:
- enables SSL verification by default
- adds command `hammer --fetch-ca-cert https://foreman.example.com` that prints server CA cert


mbacovsky:
@tstrachota I've tested this along with https://github.com/theforeman/hammer-cli-foreman/issues/293 and it works great. I appreciate the `--fetch-ce-cert` method :beer:. There are three tests failing expectations due to the passing of :verify_ssl param.

mbacovsky:
:+1: LGTM now

akofink:
I know this is an old PR, but did we add any docs for these features?

PR #236, opened 2017-04-03
Fixes #19083 - Install server CA cert without root access

mbacovsky:
--fetch-ca-cert stores the cert into ~/.hammer/certs by default.
Hammer also sets the default ssl-ca-path to there. Certificates are
stored in .pem format and the hash-named link required by OpenSSL is
created.


mbacovsky:
Output of updated `--fetch-ca-cert`:
```
hammer  --fetch-ca-cert https://my.server.com
CA certificate for my.server.com was stored to /home/mbacovsk/.hammer/certs/my.server.com.pem
Now hammer can use the downloaded certificate to verify SSL connection to the server.

Be aware that during the download hammer cannot verify whether the certificate is correct and you should verify its authenticity.

You can display certificate content with
  $ openssl x509 -text -in /home/mbacovsk/.hammer/certs/my.server.com.pem


As root you can also install the certificate and update the system-wide list of trusted CA certificates as follows:

  on Redhat systems:
  $ install /home/mbacovsk/.hammer/certs/my.server.com.pem /etc/pki/ca-trust/source/anchors/
  $ update-ca-trust

  on Debian systems:
  $ install /home/mbacovsk/.hammer/certs/my.server.com.pem /usr/local/share/ca-certificates/
  $ update-ca-certificates

```

lib/hammer_cli/ca_cert_fetcher.rb
```diff
@@ -58,6 +87,35 @@ def get_ca_cert(hostname, port)
...
+
+      # create link
+      hash = cert_hash(raw_cert)
+      cert_link = File.join(ca_path, "#{hash.to_s(16)}.%s")
+      count = 0
+      # increase hash index if file or link to different target already exist
+      while ((File.exist?(cert_link % count) || File.symlink?(cert_link % count)) &&
+          (File.symlink?(cert_link % count) == false ||
+            (File.symlink?(cert_link % count) &&
+            !(File.readlink(cert_link % count) =~ /\/#{hostname}.pem/)))) do
```
tstrachota: This clause is really difficult to read, even with the comment above. Correct me if I'm wrong but based on my assumptions that
- if `(File.exist?(some_file)` is `false` then `File.symlink?(some_file)` is `false` too
- `File.symlink?` returns only `true` or `false`

the code can be simplified to
```ruby
while File.exist?(cert_link % count) && 
      (!File.symlink?(cert_link % count) || !(File.readlink(cert_link % count) =~ /\/#{hostname}.pem/)) do
```

Maybe we could wrap the code into a function `find_link_name` or something like that.

lib/hammer_cli/ca_cert_fetcher.rb
```diff
@@ -17,7 +17,36 @@ def fetch_ca_cert(host)
...
+        puts
+        puts _("Be aware that during the download hammer cannot verify whether the certificate is correct and you should verify its authenticity.")
+        puts
+        puts _("You can display certificate content with")
+        puts "  $ openssl x509 -text -in #{cert_file}"
+        puts
+        puts
+        puts _("As root you can also install the certificate and update the system-wide list of trusted CA certificates as follows:")
+        puts
+        puts "  " + _("on Redhat systems") +  ":"
```
mbacovsky: Good idea, I'll add it

lib/hammer_cli/ca_cert_fetcher.rb
mbacovsky: Surprisingly file not exists? and is symlink? when the link points to non-existing location. Regardless I agree this needs to be simplified/refactored

tstrachota:
Feedback from @swadeley regarding the help text:
> ```
> Be aware that during the download hammer cannot verify whether the ...
> ```
> Is it relevant to say "during the download"?
> How about just "Be aware that hammer cannot verify"
> 
> ```
> You can display certificate content with ...
> ```
> You can display the  <--- add "the"

mbacovsky:
@tstrachota, I addressed all the issues mentioned above. I've also added separate commit with some refactoring, where certain parameters were turned to instance variables. What do you think?
I've also added some tests covering the link creation though I didn't found a good way to avoid the temp dir creation.

lib/hammer_cli/ca_cert_fetcher.rb
```diff
@@ -1,30 +1,68 @@
...
-        if hostname.nil?
-          $stderr.puts _("Couldn't parse URI '%s'.") % host
-          $stderr.puts scheme_error(uri)
-          return HammerCLI::EX_SOFTWARE
+        if File.directory?(rh_install_path)
+          puts cert_install_msg
+          puts "  $ install #{cert_file} #{rh_install_path}"
+          puts "  $ #{rh_update_cmd}"
         end
+        if File.directory?(deb_install_path)
```
mbacovsky: This was intentional as no root CA cert installation seems to be okay option too when there is no support for that in a system. E.g. when ca-certificates RPM package is not installed on a RH family system

tstrachota: I agree, but I don't understand how the pattern `IF rh_path_exist THEN print_rh_instructions ELSIF deb_path exist THEN print_deb_instructions` breaks it.

mbacovsky:
@tstrachota, how about changing the `create_link`, `get_ca_cert` and `store_ca_cert` visibility to public?  It could have sense to use them separately as they are providing complete functionality.
For testing the main `fetch_ca_cert` we could just stub `get_ca_cert` to return some fixture. 
Drafted in the new commit.

tstrachota:
I still think it combines two responsibilities in one class but I can live with that if we find a bit more systematic names and interfaces for the methods. I can imagine making `#store_ca_cert` and `#get_ca_cert` public. The rest would feel quite unnatural in my opinion.

`#fetch_ca_cert` would probably then be better named `#get_and_store_cert`. Also `host` should probably be moved away from the constructor. Moving all the `puts` to `.fetch_ca_cert` which is more of a helper method would make `#get_and_store_cert` more universal.

Either that or two classes (preferred from me) or some other superelegant solution you find :)

thomasmckay:
running --fetch-ca-cert put the file in ~/.hammer/cert (singular not plural certs), is that correct? The example yml shows certs. I could not get the local cert to work. My server is https://devel.example.com:50230, what should the file be in the local cert dir?

mbacovsky:
@thomasmckay, you are right, the dafault does not match the example config. Fixed (to ~/.hammer/certs).
You can also specify your own path with  --ssl-ca-path or same option in hammer config.

Hammer expects port 443 by default so your command should look like:
``` bash
hammer --fetch-ca-cert https://devel.example.com:50230
```
If it worked in the local cert dir (~/.hammer/certs) should be two files - `devel.example.com.pem` and a link to it with the hash name e.g. `2c543cd1.0`

Were you instructed to include the port to the url  by the SSL verification error?

thomasmckay:
```
$ hammer --fetch-ca-cert https://devel.example.com:50230
CA certificate for devel.example.com was stored to /home/thomasmckay/.hammer/cert/devel.example.com.pem
Now hammer can use the downloaded certificate to verify SSL connection to the server.

Be aware that hammer cannot verify whether the certificate is correct and you should verify its authenticity.

You can display the certificate content with
  $ openssl x509 -text -in /home/thomasmckay/.hammer/cert/devel.example.com.pem


As root you can also install the certificate and update the system-wide list of trusted CA certificates as follows:
  $ install /home/thomasmckay/.hammer/cert/devel.example.com.pem /etc/pki/ca-trust/source/anchors/
  $ update-ca-trust
```

the config file

```
:ssl:
  :ssl_ca_file: '/home/thomasmckay/.hammer/cert/devel.example.com.pem'
```

and then running

```
$ hammer organization list
SSL certificate verification failed
Make sure you configured the correct URL and have the server's CA certificate installed on your system.
...
```

mbacovsky:
@thomasmckay, was the cert downloaded? What is output of
```
$ openssl x509 -text -in /home/thomasmckay/.hammer/cert/devel.example.com.pem
$ ls -la  /home/thomasmckay/.hammer/cert/
```
Does the server use self-signed cert or is there any chain

mbacovsky:
@tstrachota, I split the code into separate CA certs manager and command I/O stuff. Hope you'll find it supercool.  :running_man: :poodle:

mbacovsky:
[test]

mbacovsky:
The tests are failing on file that is no longer part of the PR. It was added and then removed within this PR which is confusing Jenkins for some reason. Squashing the commits could fix the tests and can do it before merge once we agree on the changes.

thomasmckay:
I'm going to say that my setup may not be typical customers. I am running from host AAA to host BBB. AAA has /etc/hosts entry to map name CCC to BBB's ip address. BBB has configured port 1234 to redirect to a VM named CCC. If that makes sense. :) Don't hold up this PR on my comments.

My suggestion would be that the help message indicate how to specify --verify-ssl=false otherwise it will be very frustrating for users trying to get something done.

I will also note that it should be "Red Hat" not "Redhat" as the name given in the help message.

Thanks!

test/unit/ca_cert_manager_test.rb
```diff
@@ -0,0 +1,64 @@
...
+  let(:service_uri) { URI.parse("https://test.host.com") }
+  let(:ca_cert_manager) { HammerCLI::CACertManager.new(@ca_path) }
+
+  describe '#add_ca_cert' do
+    let(:cert_fixture) { File.join(File.dirname(__FILE__), '/fixtures/certs/ca_cert.pem') }
+
+    it 'fetches and stores ca cert' do
+      ca_cert_manager.stubs(:get_ca_cert).returns(File.read(cert_fixture))
+      out, err = capture_io { ca_cert_manager.add_ca_cert(service_uri) }
+      assert_equal true, File.exist?(File.join(@ca_path, "2c543cd1.0"))
```
tstrachota: Just `assert`, please

lib/hammer_cli/ca_cert_manager.rb
```diff
@@ -0,0 +1,71 @@
...
+      @ca_path = ca_path
+    end
+
+    def add_ca_cert(uri)
+      raw_cert = get_ca_cert(uri)
+      cert_file = cert_file_name(uri)
+      store_ca_cert(raw_cert, cert_file)
+    end
+
+    def get_ca_cert(uri)
```
tstrachota: Not a big thing but I would move the logic around fetching certs to a standalone class for the sake of separating responsibilities (you could then avoid stubbing `get_ca_cert` in tests). But I'm not going to block the PR if you don't like it.

test/unit/ca_cert_manager_test.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  let(:service_uri) { URI.parse("https://test.host.com") }
+  let(:ca_cert_manager) { HammerCLI::CACertManager.new(@ca_path) }
+
+  describe '#add_ca_cert' do
+    let(:cert_fixture) { File.join(File.dirname(__FILE__), '/fixtures/certs/ca_cert.pem') }
+
+    it 'fetches and stores ca cert' do
+      ca_cert_manager.stubs(:get_ca_cert).returns(File.read(cert_fixture))
+      out, err = capture_io { ca_cert_manager.add_ca_cert(service_uri) }
```
tstrachota: What output can the method produce?

lib/hammer_cli/ca_cert_fetcher.rb
```diff
@@ -1,39 +1,63 @@
...
+        puts _("Now hammer can use the downloaded certificate to verify SSL connection to the server.")
+        puts
+        puts _("Be aware that hammer cannot verify whether the certificate is correct and you should verify its authenticity.")
+        puts
+        puts _("You can display the certificate content with")
+        puts "  $ openssl x509 -text -in #{cert_file}"
+        puts
 
-        puts get_ca_cert(hostname, port)
+        cert_install_msg = "\n" + _("As root you can also install the certificate and update the system-wide list of trusted CA certificates as follows:") + "\n"
```
tstrachota: That results in printing two newlines between the `openssl` command and "As root you can also install...". Was that intentional? (just to make sure)

lib/hammer_cli/ca_cert_fetcher.rb
```diff
@@ -1,39 +1,63 @@
...
-    def self.fetch_ca_cert(host)
-      CACertFetcher.new.fetch_ca_cert(host)
-    end
-
-    def fetch_ca_cert(host)
+    def fetch_ca_cert(service_uri, ca_path)
       begin
-        uri = URI.parse(host)
+        uri = URI.parse(service_uri)
+        raise URI::InvalidURIError.new(scheme_error(uri)) unless uri.scheme == 'https'
```
tstrachota: It's probably worth checking `host.uri` here too (or in the cert manager?). When host can't be parsed it produces symlink to `~/.hammer/certs/.pem`.

```
You can display the certificate content with
  $ openssl x509 -text -in /root/.hammer/certs/.pem
```

tstrachota:
@thomasmckay I think your setup should be supported. Did you have any chance to investigate what could be wrong?

lib/hammer_cli/ca_cert_fetcher.rb
mbacovsky: @tstrachota could you share a reproducer for the message above?

lib/hammer_cli/ca_cert_manager.rb
mbacovsky: I tried to extract this code to separate calss but I ended up either with SSLSocket wrapper which seems pointless or class more difficult to use. I'm open to suggestions otherwise I'd keep it as is.

lib/hammer_cli/ca_cert_fetcher.rb
tstrachota:
```
hammer -s https:foreman.test.org host list
```

lib/hammer_cli/ca_cert_manager.rb
tstrachota: My suggestion is to remove `get_ca_cert` and `noverify_ssl_context` out of `CertManager`. Yes, it would be more or less just wrapper around `SSLSocket` and that's imho the point - get rid of the dependency. `CertManager` will only handle installing certs (string) into given location and won't care where they come from.

I find injecting values into tested objects much cleaner than stubbing methods on the tested objects (you never know what you stub out).

But if you insist on having it in one class, I'm ok. It's not any huge monolith. My only concern was that it breaks good practice;) We can refactor it later if there's need for that.

mbacovsky:
@tstrachota  I've added one more commit with requested refactoring. 
To be honest it doesn't look that bad, but it feels like we are trading design for usability here. For me it is a bit on the over-engineered side. We avoided stubbing the method but only because we do not test CaCertFetcher. 
Anyway I'm okay with it both ways, so feel free to pick the one you like more.

Let me know if there are any other objections. In the meantime I'll try to reproduce the issue @thomasmckay hit. I believe that the reason for the failure was the root cert didn't have CA:TRUE constraint.

test/unit/ca_cert_manager_test.rb
```diff
@@ -0,0 +1,64 @@
...
+      FileUtils.touch(cert_file)
+      assert File.exist?(File.join(@ca_path, "#{hash_file}.0"))
+    end
+
+    it "does not create new link if it already exist" do
+      FileUtils.touch(cert_file)
+      File.symlink(cert_file, File.join(@ca_path, "#{hash_file}.0"))
+      ca_cert_manager.create_link(hash, cert_file)
+      assert File.exist?(File.join(@ca_path, "#{hash_file}.0"))
+      assert_equal false, File.exist?(File.join(@ca_path, "#{hash_file}.1"))
```
tstrachota: Btw you can use `refute` here

tstrachota:
@mbacovsky thanks, looks ok now. If the "worse usability" means a developer needs to write few more characters to use it but the code is more flexible, I'm perfectly fine with that. Btw you really wouldn't need to stub anything even when testing `CaCertFetcher` if the instances were correctly injected (with meaningful defaults).

:+1: pending investigation of the failure from the previous comment

mbacovsky:
@thomasmckay I investigated a bit the cert issue you've reported and found out [1](the accepted answer) that it is possible to sign your certificates with ca that has CA:FALSE but it is not possible to verify them against such ca. From the information I have it seems that your CA does not have CA:TRUE and the reported behavior is then expected.

To make sure --fetch-ca-cert downloads the cert correctly we can compare the `/home/thomasmckay/.hammer/cert/devel.example.com.pem` with the CA cert on the server or with the output of
```
openssl s_client -showcerts -connect devel.example.com.pem:50230 </dev/null
```

Are you able to verify the connection with the CA cert manualy downloaded?
```
hammer --ssl-ca-file <cacert> organization list
```
or
```
curl --cacert <cacert> -v https://devel.example.com.pem:50230
```
[1] https://superuser.com/questions/462295/openssl-ca-and-non-ca-certificate

mbacovsky:
@tstrachota I tried to move the cert download to the manager and I like it more. Added the dependency injection and test. Is it acceptable?
I've never heard of `refute`, nice,  updated.

tstrachota:
@mbacovsky either one works for me. I find the previous implementation better, but that's matter of personal preference rather than some tangible advantages.

Your comment about CA:FALSE makes sense. So :+1: to merge regardless of whether you decide to keep your last change in manager or not.

mbacovsky:
I'll remove the last commit and squash to make the tests green again.

mbacovsky:
@tstrachota, updated, squashed, tests are green. Could you approve your change request to have everything green. I'll merge it then.

PR #238, opened 2017-05-02
Fixes #19431 - retry command on session expiry

tstrachota:
Generic mechanism for retrying commands when they return `EX_RETRY` either from method `execute` or from exception handler.

### DIFF
--- a/lib/hammer_cli/exit_codes.rb
+++ b/lib/hammer_cli/exit_codes.rb
@@ -4,7 +4,7 @@ module HammerCLI
   EX_USAGE        = 64      # command line usage error
   EX_DATAERR      = 65      # data format error
   EX_NOINPUT      = 66      # cannot open input
-  EX_NOUSER       = 67      # addressee unknown 
+  EX_NOUSER       = 67      # addressee unknown
   EX_NOHOST       = 68      # host name unknown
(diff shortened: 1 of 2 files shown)


tstrachota:
Requires https://github.com/theforeman/hammer-cli-foreman/pull/309 to work for session expiry.

tstrachota:
@ares @mbacovsky I guess you would be interested in taking a look at the code.

lib/hammer_cli/exit_codes.rb
```diff
@@ -20,4 +20,5 @@ module HammerCLI
   # non POSIX codes
   EX_NOT_FOUND    = 128     # resource was not found
   EX_UNAUTHORIZED = 129     # authorization failed
+  EX_RETRY        = 666     # retry the command
```
ares: what a number :-)

ares:
Works very well :+1: Few ideas for later improvements - it might not log the unauhtorized 401 as error log level when ran with `-d`, especially the trace indicates some problem, even though 4xx HTTP code means it's client error. I can also imagine it would suggest the login from last time so when I hit enter on username prompt it would use the previous login, but those are really just nitpick. +1000 for this super cool feature, could we get this in some upcoming release? :-)

tstrachota:
Merged, thanks for testing it @ares

PR #240, opened 2017-05-05
refs #19083 - interpolate variables in translations correctly

domcleal:


tstrachota:
OMG how could we miss it?! :+1: Thank you @domcleal !

PR #241, opened 2017-05-17
Fixes #19573 - Make sure assert_match is called in correct context

tstrachota:


mbacovsky:
:+1: Looks sane, fixes the issue and does not break anything :meat_on_bone: :poodle:

PR #244, opened 2017-05-26
Fixes #19676 - saving defaults from hammer shell

tstrachota:
See the linked redmine issue for reproducer steps.

### DIFF
--- a/lib/hammer_cli/defaults.rb
+++ b/lib/hammer_cli/defaults.rb
@@ -23,20 +23,18 @@ module HammerCLI
     end
 
     def delete_default_from_conf(param)
-      update_defaults_file do |defaults|
-        defaults.delete_if { |k,| defaults_match?(k, param) }
-      end
+      @defaults_settings.delete_if { |k,| defaults_match?(k, param) }
+      update_defaults_file
     end
 
     def add_defaults_to_conf(default_options, provider)
       create_default_file if defaults_settings.empty?
-      update_defaults_file do |defaults|
-        default_options.each do |key, value|
-          key = switch_to_name(key).to_sym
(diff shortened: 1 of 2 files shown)


mbacovsky:
@tstrachota, the patch looks good. But I'm not sure it fixes the whole problem. When updating defaults in the hammer shell, the hammer instance defaults are not updated. Just the config file. 
```
hammer> defaults list
----------------|---------------------
PARAMETER       | VALUE               
----------------|---------------------
----------------|---------------------

hammer> defaults add --param-name location-id --provider foreman
Added location-id default-option with value that will be generated from the server.

hammer> defaults list
----------------|---------------------
PARAMETER       | VALUE               
----------------|---------------------
----------------|---------------------
```
The config looks good now and after reloading the shell all is right.
```
$ cat ~/.hammer/defaults.yml 
---
:defaults:
  :location-id:
    :provider: foreman
```

The way how the `update_defaults_file` works is weird. It loads the config, update it and store regardless what the instance content is (settings can be provided in constructor so it can be completely different from the config file content). While the defaults instance in the shell persists it is not reloaded and miss the new data. I'd suggest to change the `update_defaults_file`  to update `@defaults_settings` and overwrite the config file with it. What do you think?

tstrachota:
@mbacovsky added a commit that fixes behavior in shell. Should be squashed before merge.

ares:
Fixes the problem for me, code looks good to me

tstrachota:
@mbacovsky I've addressed your comments, can you please re-review?

mbacovsky:
:+1: Thanks @tstrachota, :hammer:  :shell: works great now!

PR #246, opened 2017-06-22
Refs #17135 - Replace CSV parser in List Normalizer

mbacovsky:
The ruby CSV parser causes some usability problems when used to parse list options for more details see [1].
The new parser should be more flexible in syntax and closer to shell quoting conventions. 

[1] https://github.com/theforeman/hammer-cli-foreman/pull/312

### DIFF
--- a/lib/hammer_cli/csv_parser.rb
+++ b/lib/hammer_cli/csv_parser.rb
@@ -1,0 +1,74 @@
+module HammerCLI
+  class CSVParser
+
+    def initialize
+      reset_parser
+    end
+
+    def parse(data)
+      return [] if data.nil?
+      reset_parser
+      data.split('').each do |char|
+        handle_escape(char) || handle_quoting(char) || handle_comma(char) || add_to_buffer(char)
+      end
+      raise ArgumentError.new(_("Illegal quoting in %{buffer}") % { :buffer => @buffer }) unless @last_quote.nil?
+      clean_buffer
+      @value
+    end
+
+    private
+
+    def handle_comma(char)
+      if char == ','
+        clean_buffer
+        true
+      else
+        false
+      end
+    end
+
+    def handle_quoting(char)
+      if @last_quote.nil? && ["'", '"'].include?(char)
+        @last_quote = char
+        true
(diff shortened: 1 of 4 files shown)


mbacovsky:
Examples of the required CSV syntax can be seen in the tests. I'd be glad for suggestions on adding other inputs to the tests.

orrabin:
I tried this with  theforeman/hammer-cli-foreman#312 all these options worked correctly for me:
`--override-value-order 'domain,os'`
`--override-value-order "'domain,os', hostgroup"`
`--override-value-order "'domain,os', 'hostgroup'"`
`--override-value-order "'domain,os', 'hostgroup, os', fqdn"`

lib/hammer_cli/csv_parser.rb
```diff
@@ -0,0 +1,74 @@
...
+  class CSVParser
+
+    def initialize
+      reset_parser
+    end
+
+    def parse(data)
+      return [] if data.nil?
+      reset_parser
+      data.split('').each do |char|
```
tstrachota: Why not `each_char`?

test/unit/csv_parser_test.rb
```diff
@@ -0,0 +1,53 @@
...
+    it "parses a comma separated string with values including comma" do
+      parser.parse('a,b,"c,d"').must_equal ['a', 'b', 'c,d']
+    end
+
+    it "parses a comma separated string with values including comma (dquotes)" do
+      parser.parse("a,b,'c,d'").must_equal ['a', 'b', 'c,d']
+    end
+
+    it "raises quoting error" do
+      proc { parser.parse('1,"3,4""s') }.must_raise ArgumentError
```
tstrachota: Thank you for adding the syntax error reporting! Can you test the error message here, please?
Parsing `'1,"3,4""s'` produces `Illegal quoting in 3,4s`. Shouldn't we pass the whole original string? 
(in this case `Illegal quoting in "3,4""s`)

mbacovsky:
@tstrachota, updated

tstrachota:
:+1: thanks @mbacovsky

PR #247, opened 2017-07-25
Fixes #20181 - add full-help command

tstrachota:
Moving full-help command from https://github.com/tstrachota/hammer-cli-experimental/ into the core.


ares:
works just fine :+1:

PR #248, opened 2017-08-04
Refs #12401 - More detailed description of ssl options

tstrachota:


config/cli_config.template.yml
```diff
@@ -34,23 +34,25 @@
 
 # SSL auth options
 #:ssl:
-  # Path to a ca file
+  # Path to a CA file in PEM format, mutually exclusive with :ssl_ca_path:
   #:ssl_ca_file: '/path/to/ca_certificate.pem'
 
-  # Path to a direcotry with ca file
```
mbacovsky: May be worth mentioning that the directory must follow some rules as mentioned in OpenSSL docs. Took me some time to figure out:
```
The directory should contain one certificate or CRL per file in PEM format, with a file name of the form hash.N
 for a certificate, or hash.rN for a CRL. The hash is the value returned by the X509_NAME_hash function applied
 to the subject name for certificates or issuer name for CRLs. The hash can also be obtained via the -hash option 
of the x509 or crl commands.
```
https://www.openssl.org/docs/man1.1.0/crypto/X509_LOOKUP_hash_dir.html see `Hashed Directory Method` there.
Shorter explanation is in https://stackoverflow.com/questions/38529918/ssl-ca-path-with-iosocketssl-doesnt-use-certs-in-directory/38539719

mbacovsky:
LGTM. The extra comments are useful. I had some difficulties to make the `:ssl_file_path` working so it may be useful to add some hint as mentioned inline. What do you think?

tstrachota:
@mbacovsky makes sense, updated.

mbacovsky:
:+1: @tstrachota thanks, looks good now. :book: :books: :arrow_up:

PR #249, opened 2017-08-07
Fixes #20513 - shareable locale Makefile and Rakefile definitions

tstrachota:
This change will allow plugins to share common parts of rake tasks and make targets around extracting strings for translation.

### DIFF
--- a/doc/i18n.md
+++ b/doc/i18n.md
@@ -46,7 +46,6 @@ HammerCLI::I18n.add_domain(HammerCLIAwesomePlugin::I18n::LocaleDomain.new)
 ```
 
 Then you have to export strings, translate them and place the files in a directory structure.
-Inspiration on how to export the translations can be found in hammer's [Rakefile](../Rakefile).
 
 Typical directory structure for translation files look like this:
 ```
@@ -61,6 +60,29 @@ locale
 │       └── hammer-cli-awesome.mo
 └── hammer-cli-awesome.pot
 ```
+
+You can re-use Rake tasks and Makefile targets for extracting translations and integration with transifex that hammer provides.
+To do that, add following lines to the plugin's Rakefile:
+```ruby
+require "hammer_cli_awesome/version"
+require "hammer_cli_awesome/i18n"
+require "hammer_cli/i18n/find_task"
+HammerCLI::I18n::FindTask.define(HammerCLIAwesome::I18n::LocaleDomain.new, HammerCLIAwesome.version)
+```
+
+and create `locale/Makefile` with following content:
+```make
+DOMAIN = hammer-cli-awesome
+VERSION = $(shell bundle exec ruby -e 'require "rubygems"; spec = Gem::Specification::load("../hammer_cli_awesome.gemspec"); puts spec.version')
+MAIN_MAKEFILE = $(shell bundle exec ruby -e 'require "hammer_cli"; puts HammerCLI::I18n.main_makefile')
+
+include $(MAIN_MAKEFILE)
+```
+
+Make sure you have a project created in [transifex](www.transifex.com) and correct configuration stored in your plugins repository ([example config file](../.tx/config)).
+
+
+Calling `make -C ./locale tx-update` will then extract new strings, update `.po` and `.mo` files and commit the changes.
 
 
 ### Translation tips
--- a/locale/Makefile
+++ b/locale/Makefile
@@ -1,66 +1,10 @@
 #
 # Makefile for PO merging and MO generation. More info in the README.
 #
-# make all-mo (default) - generate MO files
-# make check - check translations using translate-tool
-# make tx-update - download and merge translations from Transifex
-# make clean - clean everything
+# Run make help to list available targets
 #
 DOMAIN = hammer-cli
-VERSION = $(shell ruby -e 'require "rubygems";spec = Gem::Specification::load("../hammer_cli.gemspec");puts spec.version')
-POTFILE = $(DOMAIN).pot
-MOFILE = $(DOMAIN).mo
-POFILES = $(shell find . -name '$(DOMAIN).po')
-MOFILES = $(patsubst %.po,%.mo,$(POFILES))
-POXFILES = $(patsubst %.po,%.pox,$(POFILES))
-EDITFILES = $(patsubst %.po,%.edit.po,$(POFILES))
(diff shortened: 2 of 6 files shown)


tstrachota:
@mbacovsky could you please to the review?

doc/i18n.md
```diff
@@ -62,6 +61,29 @@ locale
...
+VERSION = $(shell ruby -e 'require "rubygems"; spec = Gem::Specification::load("../hammer_cli_awesome.gemspec"); puts spec.version')
+MAIN_MAKEFILE = $(shell ruby -e 'require "hammer_cli"; puts HammerCLI::I18n.main_makefile')
+
+include $(MAIN_MAKEFILE)
+```
+
+Make sure you have a project created in [transifex](www.transifex.com) and correct configuration stored in your plugins repository ([example config file](../.tx/config)).
+
+
+Calling `make -C ./locale tx-update` will then extract new strings, update `.po` and `.mo` files and commit the changes.
```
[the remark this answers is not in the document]
tstrachota: I added a sentence that warns developers. Using `bundle exec` is imho correct and standard way. What was your idea of finding the main makefile location without using bundler?

tstrachota:
Updated

tstrachota:
[test]

doc/i18n.md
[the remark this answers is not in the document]
mbacovsky: The point was it was not necessary in the past to use bundler to invoke the make. From the users perspective the make suddenly fails with backtrace if used as usual. I'd expect Makefile to handle the situation and use bundler as a fallback option on its own (as `make` usually does in other projects).

locale/Makefile
```diff
@@ -7,60 +7,7 @@
...
-VERSION = $(shell ruby -e 'require "rubygems";spec = Gem::Specification::load("../hammer_cli.gemspec");puts spec.version')
-POTFILE = $(DOMAIN).pot
-MOFILE = $(DOMAIN).mo
-POFILES = $(shell find . -name '$(DOMAIN).po')
-MOFILES = $(patsubst %.po,%.mo,$(POFILES))
-POXFILES = $(patsubst %.po,%.pox,$(POFILES))
-EDITFILES = $(patsubst %.po,%.edit.po,$(POFILES))
-TIMESTAMPFILES = $(patsubst %.po,%.po.time_stamp,$(POFILES))
+VERSION = $(shell ruby -e 'require "rubygems"; spec = Gem::Specification::load("../hammer_cli.gemspec"); puts spec.version')
+MAIN_MAKEFILE = $(shell ruby -e 'require "hammer_cli"; puts HammerCLI::I18n.main_makefile')
```
mbacovsky: @tstrachota I found out we use `bundle exec` already in the `extract-strings` target so I think it makes sense to include it here as well. It is pointless to expect calling any of the targets outside of devel env anyway, isn't it?

locale/Makefile.def
```diff
@@ -0,0 +1,67 @@
...
+#
+
+POTFILE = $(DOMAIN).pot
+MOFILE = $(DOMAIN).mo
+POFILES = $(shell find . -name '$(DOMAIN).po')
+MOFILES = $(patsubst %.po,%.mo,$(POFILES))
+POXFILES = $(patsubst %.po,%.pox,$(POFILES))
+EDITFILES = $(patsubst %.po,%.edit.po,$(POFILES))
+TIMESTAMPFILES = $(patsubst %.po,%.po.time_stamp,$(POFILES))
+
```
mbacovsky: How about adding help target?
```bash
.PHONY: help

help: ## This help message
	@echo -e "$$(grep -hE '^\S+:.*##' $(MAKEFILE_LIST) | sed -e 's/:.*##\s*/:/' -e 's/^\(.\+\):\(.*\)/\\x1b[36m\1\\x1b[m:\2/' | column -c2 -t -s :)"
```
Then commenting the target with double hash like: `clean: ## Clean everything` would produce:

```
$ make help
help   This help message
...
check  Check translations using translate-tool
clean  Clean everything
```
Seems useful. I've seen it adopted by some other projects

locale/Makefile.def
```diff
@@ -0,0 +1,67 @@
...
+%.mo: %.po
+	mkdir -p $(shell dirname $@)/LC_MESSAGES
+	msgfmt -o $(shell dirname $@)/LC_MESSAGES/$(MOFILE) $<
+
+# Generate MO files from PO files
+all-mo: $(MOFILES)
```
mbacovsky: As this is going to be shared and we no longer don't have control over what is in the project dirs I'd suggest to add .PHONY for non-file targets. Note: Makefile style guide suggests to add `.PHONY: ` immediately before the target rather then listing all at one place.
```
.PHONY: all-mo
```

locale/Makefile.def
tstrachota: Nice, I added that.

tstrachota:
@mbacovsky updated. I added `bundle exec` to the commands. It's true that we're already using it in `extract-strings` target.

locale/Makefile.def
```diff
@@ -0,0 +1,77 @@
...
+extract-strings:
+	bundle exec rake gettext:find
+
+.PHONY: tx-update
+tx-update: extract-strings tx-pull update-po $(MOFILES) ## Download and merge translations from Transifex
+	git add ../locale
+	git commit -m "i18n - extracting new, pulling from tx"
+	@echo Changes commited!
+
+.PHONY:clean
```
mbacovsky: nitpick - space missing

locale/Makefile.def
```diff
@@ -0,0 +1,77 @@
...
+# Merge .edit.po into .po
+.PHONY: update-po
+update-po:
+	for f in $(EDITFILES); do \
+		msgcat --use-first --no-location $$f $${f//.edit/} > $${f//.edit/}; \
+	done
+
+# Extract strings and update the .pot, prepare .edit.po files
+.PHONY: extract-strings
+extract-strings:
```
mbacovsky: @tstrachota, were some of the target not documented/added to the help on purpose?

tstrachota: Yes, I intentionally documented only the main targets to keep the help simple and left out targets that are rather "intermediate".

tstrachota:
Updated

mbacovsky:
Thanks, @tstrachota :banana: :gorilla: 
:+1:

PR #250, opened 2017-08-10
Fixes #19511 - Consistant capitalization in hammer help

rahulbajaj0509:


tstrachota:
Works fine, thanks @rahulbajaj0509 
Could you please add one more test that would check solely that adding an option with lower case description is capitalized?

tstrachota:
Thanks for adding the test!

It discovered one more thing: I think that we should keep strings that are all capital as they are. It's very difficult to introduce such strings by mistake because a reviewer usually easily spots it. This makes me think that if such string appears it's by intention and we should not change it.

Therefore I would:
`alfa option` -> `Alfa option`
`BETA option` -> `BETA option`

What do you think?

Apart from that the code and the test look good so far.

rahulbajaj0509:
@tstrachota i feel we can do something like this : 
` s = "foreman provides provisioning" `
 `s[0].upcase + s[1..(s.length - 1)] `

but this will increase the time complexity i guess, so imho  we should go with capitalize method :)

tstrachota:
It could be simplified to `s[0].upcase + s[1..-1]`. Or how about `s.gsub(/^(.)/) { $1.capitalize }` ?
I think that time complexity is not a big deal here because the method will be executed only for printing help, not elsewhere.

Another thing is that we should be using `Unicode::capitalize` to make sure languages with unicode characters are capitalized properly. But I'm not sure it's worth adding a dependency on `unicode` gem.

rahulbajaj0509:
@tstrachota agreed 👍  i shall  make the changes.

tstrachota:
OK, let's use `unicode`. I did some basic performance tests and using it doesn't seem to have dramatically negative impact on performance.

rahulbajaj0509:
@tstrachota using `Unicode::capitalize`, it will again lead to https://github.com/theforeman/hammer-cli/pull/250#issuecomment-321572110 
As the capitalise function works like this : 
`irb(main):002:0> Unicode::capitalize("BLOG")
=> "Blog"`

tstrachota:
Then we probably need to combine it into `s.gsub(/^(.)/) { Unicode::capitalize($1) }`

tstrachota:
:+1: thanks @rahulbajaj0509

PR #252, opened 2017-09-19
Fixes #21013 - Number normalizers for Integer params

mbacovsky:
Option builder was updated to recognize params of type Integer.
During investigation of this it was found out that apipie sends
expected_type 'numeric' and not 'number' (since v. 0.0.6). Fixed
that as well.


tstrachota:
:+1: thanks @mbacovsky

akofink:
For the record, this is a good example of the type of change I would like to know about. :hammer:

mbacovsky:
@akofink sorry for that, this seemed to be cosmetic change as I believe the API converts the types anyway.

PR #253, opened 2017-10-09
Fixes #20978 - Allow hiding headers from output of list actions

ofedoren:


tstrachota:
[test]

lib/hammer_cli/main.rb
```diff
@@ -39,7 +39,7 @@ class MainCommand < AbstractCommand
     option ["--interactive"], "INTERACTIVE", _("Explicitly turn interactive mode on/off"),
       :format => HammerCLI::Options::Normalizers::Bool.new,
       :context_target => :interactive
-
+    option ["--no-headers"], :flag, _("Hide headers from output")
```
mbacovsky: The option should also be defined in `bin/hammer` (like e.g.https://github.com/theforeman/hammer-cli/blob/master/bin/hammer#L31) otherwise the pre-parser there fail on unknown option and will not do its job:
```
$  hammer --no-headers -s https://fqdn  hostgroup list
Could not load the API description from the server: 
Failed to open TCP connection to foreman.my.lan:80 (getaddrinfo: Name or service not known)
  - is the server down?
```

ofedoren: Updated

mbacovsky:
[test]

mbacovsky:
@ofedoren, looks good now. 
I checked the Redmine issue and found there a request to support setting this also in hammer config.

I'd expect it in the `:ui` section in https://github.com/theforeman/hammer-cli/blob/master/config/cli_config.template.yml#L10. See here https://github.com/theforeman/hammer-cli/blob/ef9ac981e54fbc5fd716b4b4243265bc980296c8/bin/hammer#L83 how the CLI option gets to the settings. https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/context.rb#L5 might be a good place to get it into the context.

ofedoren:
@mbacovsky, done.

tstrachota:
@mbacovsky ping. What's the status here?

mbacovsky:
[test]

mbacovsky:
Works good. Looks good. Well done, thanks a lot @ofedoren! 
I've re-run the test, will merge when it gets green.

PR #256, opened 2017-10-24
Fixes #20607 - Allow hiding fields which are missing from api output

ofedoren:
Related to https://github.com/theforeman/hammer-cli-foreman/pull/345

### DIFF
--- a/lib/hammer_cli/output/adapter/abstract.rb
+++ b/lib/hammer_cli/output/adapter/abstract.rb
@@ -50,12 +50,13 @@ module HammerCLI::Output::Adapter
       path = field.path
 
       path.inject(record) do |record, path_key|
-        if record && record.kind_of?(Hash) && record.has_key?(path_key.to_sym)
+        return nil unless record && record.is_a?(Hash)
+        if record.key?(path_key.to_sym)
           record[path_key.to_sym]
-        elsif record && record.kind_of?(Hash) && record.has_key?(path_key.to_s)
+        elsif record.key?(path_key.to_s)
           record[path_key.to_s]
         else
-          return nil
+          HammerCLI::Output::DataMissing.new
         end
       end
     end
@@ -67,6 +68,17 @@ module HammerCLI::Output::Adapter
     def output_stream
       return @context[:output_file] if @context.has_key?(:output_file)
       $stdout
+    end
+
+    def displayable_fields(fields, record, compact_only: false)
+      fields.select do |field|
+        field_data = data_for_field(field, record)
+        if compact_only && !field_data.is_a?(HammerCLI::Output::DataMissing)
+          true
+        else
+          field.display?(field_data)
+        end
+      end
     end
 
     private
--- a/lib/hammer_cli/output/adapter/base.rb
+++ b/lib/hammer_cli/output/adapter/base.rb
@@ -27,17 +27,11 @@ module HammerCLI::Output::Adapter
       HammerCLI::Output::FieldFilter.new(filtered)
     end
 
-    def filter_fields(fields, data)
-      field_filter.filter(fields).reject do |field|
-        field_data = data_for_field(field, data)
-        not field.display?(field_data)
-      end
-    end
-
     def render_fields(fields, data)
       output = ""
 
-      fields = filter_fields(fields, data)
+      fields = field_filter.filter(fields)
+      fields = displayable_fields(fields, data)
 
       label_width = label_width(fields)
 
--- a/lib/hammer_cli/output/adapter/csv.rb
+++ b/lib/hammer_cli/output/adapter/csv.rb
@@ -157,6 +157,7 @@ module HammerCLI::Output::Adapter
     end
 
     def print_collection(fields, collection)
+      fields = displayable_fields(fields, collection.first, compact_only: true)
       rows = row_data(fields, collection)
       # get headers using columns heuristic
       headers = rows.map{ |r| Cell.headers(r, @context) }.max_by{ |headers| headers.size }
--- a/lib/hammer_cli/output/adapter/table.rb
+++ b/lib/hammer_cli/output/adapter/table.rb
@@ -22,6 +22,7 @@ module HammerCLI::Output::Adapter
 
     def print_collection(all_fields, collection)
       fields = field_filter.filter(all_fields)
+      fields = displayable_fields(fields, collection.first, compact_only: true)
 
       formatted_collection = format_values(fields, collection)
       # calculate hash of column widths (label -> width)
@@ -52,7 +53,7 @@ module HammerCLI::Output::Adapter
       # and there is no --no-headers option
       output_stream.puts line unless formatted_collection.empty? || @context[:no_headers]
 
-      if collection.meta.pagination_set? && collection.count < collection.meta.subtotal
+      if collection.respond_to?(:meta) && collection.meta.pagination_set? && collection.count < collection.meta.subtotal
         pages = (collection.meta.subtotal.to_f/collection.meta.per_page).ceil
         puts _("Page %{page} of %{total} (use --page and --per-page for navigation).") % {:page => collection.meta.page, :total => pages}
       end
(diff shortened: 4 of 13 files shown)


mbacovsky:
[test]

tstrachota:
[test]

mbacovsky:
[test]

ofedoren:
@mbacovsky  , @tstrachota updated.

tstrachota:
[test]

lib/hammer_cli/output/adapter/abstract.rb
```diff
@@ -55,13 +55,24 @@ def self.data_for_field(field, record)
...
+          # there is no way to recognize why nil is returned from this
+          # method. Raising signals that there is missing response from server,
+          # not just nil value is returned.
+          raise HammerCLI::MissingAPIError
         end
       end
     end
 
     def data_for_field(field, record)
-      Abstract.data_for_field(field, record)
```
tstrachota: Btw you can use rescue without the additional begin/end block:
```ruby
def data_for_field(field, record)
  Abstract.data_for_field(field, record)
rescue HammerCLI::MissingAPIError
  # ...
end
```

lib/hammer_cli/output/adapter/abstract.rb
```diff
@@ -55,13 +55,24 @@ def self.data_for_field(field, record)
...
+      begin
+        return Abstract.data_for_field(field, record)
+      rescue HammerCLI::MissingAPIError
+        if field.hide_missing? && (self.class != CSValues)
+          raise HammerCLI::MissingAPIError
```
tstrachota: `rescue HammerCLI::MissingAPIError => e` and then `raise e`. Btw why is `#dafa_for_field` implemented differently from `.data_for_field` ? Shouldn't they behave the same?

lib/hammer_cli/exceptions.rb
```diff
@@ -5,5 +5,6 @@ class OperationNotSupportedError < StandardError; end
   class ModuleLoadingError < StandardError; end
   class ModuleDisabledButRequired < StandardError; end
   class NoCACertificate < StandardError; end
+  class MissingAPIError < StandardError; end
```
tstrachota: The error name is a bit misleading. Could you rename it to something like `DataKeyMissingError`?

lib/hammer_cli/output/adapter/csv.rb
```diff
@@ -83,7 +83,11 @@ def self.expand_container(field, data, formatters)
       end
 
       def self.data_for_field(field, data)
-        HammerCLI::Output::Adapter::CSValues.data_for_field(field, data)
+        begin
+          return HammerCLI::Output::Adapter::CSValues.data_for_field(field, data)
+        rescue HammerCLI::MissingAPIError
+          return nil
```
tstrachota: Same here. Why is the class implementation different from the instance one?

ofedoren:
@tstrachota, updated. 

Thank you for suggesting a better solution. https://github.com/theforeman/hammer-cli-foreman/pull/345 is redundant now and should be closed I guess.

lib/hammer_cli/output/adapter/abstract.rb
```diff
@@ -69,6 +70,13 @@ def output_stream
       $stdout
     end
 
+    def displayable_fields(fields, record)
```
tstrachota: Nice, thank you for extracting it to a method.

I know it's not your code, but since you're touching the area... I noticed that there's some redundancy in the **base adapter**:
https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/output/adapter/base.rb#L31-L34
https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/output/adapter/base.rb#L47
and in the **tree adapter**:
https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/output/adapter/tree_structure.rb#L23-L26
https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/output/adapter/tree_structure.rb#L34

Both adapters first filter out fields that aren't displayable and checks for the same thing later again. I think that we could either use your new method to filter them out once for good in `filter_fields` or remove that `reject` block from filter_fields. The latter is more effective because we wouldn't call `data_for_field` twice.
What do you think?

lib/hammer_cli/output/adapter/table.rb
```diff
@@ -52,7 +53,7 @@ def print_collection(all_fields, collection)
       # and there is no --no-headers option
       output_stream.puts line unless formatted_collection.empty? || @context[:no_headers]
 
-      if collection.meta.pagination_set? && collection.count < collection.meta.subtotal
+      if collection.respond_to?(:meta) && collection.meta.pagination_set? && collection.count < collection.meta.subtotal
```
tstrachota: Thank you for fixing this!

lib/hammer_cli/output/fields.rb
```diff
@@ -18,11 +19,15 @@ def hide_blank?
       @hide_blank
     end
 
+    def hide_missing?
```
tstrachota: Nitpick: could be `attr_reader :hide_missing?`

lib/hammer_cli/output/adapter/abstract.rb
tstrachota: > And let the field_filter decide which fields will be kept for rendering.

Ack, I don't care that much about which of the ways you choose.

> Also I'm not sure now if displayable_fields method should be in adapter context instead of fields filter.

I was thinking about the same. In future, we'll most likely need to detect what fields are requested to be displayed prior to fetching data from API to conditionally load more/less data (eg. when GrephQL is used). On the other hand the "data for field" aren't known prior to the API request so we'll have to split the logic.
Anyway, I think that we should do this refactoring in a separate PR. This patch is nice and simple improvement that doesn't block us from doing it later. So, let's go step by step and use your `displayable_fields` for now.

ofedoren:
@tstrachota, updated.

Since ```displayable_fields``` is used in **csv** and **table** adapters, I refactored other adapters in the same manner as well.

Also, because [hide_missing?](https://github.com/ofedoren/hammer-cli/blob/b209d202a37b63deaecbe8ae1f69cab348aec0aa/lib/hammer_cli/output/fields.rb#L22) returns a boolean value, I'd like to keep as it is, since making ```attr_reader``` with question mark is not allowed in ruby.

lib/hammer_cli/output/adapter/base.rb
```diff
@@ -13,6 +13,8 @@ def print_record(fields, record)
     end
 
     def print_collection(fields, collection)
+      fields = field_filter.filter(fields)
+      fields = displayable_fields(fields, collection.first)
```
tstrachota: It was missing on purpose. This functionality is disabled for table and csv adapters because we can't tell what data are in the other records. Using the first record for hiding fields that are missing would be probably fair guess, but the same method can remove also fields that are blank (`""`). That could result in some false positives -> whole column would be hidden just because the first record contained empty string.

lib/hammer_cli/output/adapter/csv.rb
```diff
@@ -157,6 +157,7 @@ def print_record(fields, record)
     end
 
     def print_collection(fields, collection)
+      fields = displayable_fields(fields, collection.first)
```
tstrachota: Similarly to my note in the base adapter. We shouldn't filter fields in printing collections in csv.

lib/hammer_cli/output/adapter/tree_structure.rb
```diff
@@ -7,33 +7,28 @@ def initialize(context={}, formatters={})
...
-        not field.display?(field_data)
-      end
-    end
-
     def render_fields(fields, data)
-      fields = filter_fields(fields, data)
-
-      fields.reduce({}) do |hash, field|
+      rendered = {}
+      fields.each do |field|
```
tstrachota: What was the reason for translating reduce into more verbose each block?

ofedoren:
@tstrachota, updated.

I changed ```displayable_fields``` method, so **csv** and **table** adapters will not display field only when data is missing for this field in the first record (it does not remove field if data is just blank and I assume if data is missing in the first record, then it's missing in the others as well).

tstrachota:
Works as expected :champagne: . Thank you @ofedoren !

PR #259, opened 2017-11-14
Fixes #21627 - handle 400 bad request - concat response message

swapab:
This is an attempt to handle all 400 bad request and shows a user-friendly message based on HTTP response.


mbacovsky:
@swapab, thanks for the patch. Unfortunately this should go to `hammer-cli-foreman` repo (https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/exception_handler.rb)
The reason is the response format is Foreman specific and thus should not be in the core which we want to keep as generic as possible. Would you mind moving your PR there? 

The code looks good otherwise. I appreciate the tests. We have no pattern for introducing new exit codes in `hammer-cli-forean` yet but you can use existing `HammerCLI::EX_DATAERR` if you wish.

swapab:
@mbacovsky That does make sense.
Moved it to https://github.com/theforeman/hammer-cli-foreman/pull/341

PR #260, opened 2017-11-15
Fixes #17252 - Allow empty parameters in hammer

mbacovsky:
This change allows user to use nil as a valid value for an argument.

  $ hammer command --option=NIL
  $ HAMMER_NIL=NULL hammer command --option=NULL

Just before option normalization the 'NIL' string is replaced for
special internal object which help us to distinguish between unused
options and options set to nil. The internal object is replaced with
nill fter all options are collected and unused options filtered out.

ToDo:
- [ ] ~~test with searchables~~
- [X] docs
- [X] colision with defaults

### DIFF
--- a/lib/hammer_cli/options/option_collector.rb
+++ b/lib/hammer_cli/options/option_collector.rb
@@ -19,7 +19,7 @@ module HammerCLI
       end
 
       def options
-        @options ||= all_options.reject {|key, value| value.nil? && all_options_raw[key].nil? }
+        @options ||= all_options.reject { |key, value| value.nil? && all_options_raw[key].nil? }
       end
       
       private
--- a/lib/hammer_cli/options/option_collector.rb
+++ b/lib/hammer_cli/options/option_collector.rb
@@ -29,7 +29,7 @@ module HammerCLI
       end
       
       def translate_nil(value)
-        value == HammerCLI::NIL ? nil : value
+        value == HammerCLI::NilValue ? nil : value
       end
     end
   end
--- a/lib/hammer_cli/options/option_collector.rb
+++ b/lib/hammer_cli/options/option_collector.rb
@@ -8,14 +8,28 @@ module HammerCLI
         @option_sources = option_sources
       end
 
-      def all_options
-        @all_options ||= @option_sources.inject({}) do |all_options, source|
+      def all_options_raw
+        @all_options_raw ||= @option_sources.inject({}) do |all_options, source|
           source.get_options(@recognised_options, all_options)
         end
       end
 
+      def all_options
+        @all_options ||= translate_nils(all_options_raw)
+      end
+
       def options
-        @options ||= all_options.reject {|key, value| value.nil? }
+        @options ||= all_options.reject {|key, value| value.nil? && all_options_raw[key].nil? }
+      end
+      
+      private
+      
+      def translate_nils(opts)
+        Hash[ opts.map { |k,v| [k, translate_nil(v)] } ]
+      end
+      
+      def translate_nil(value)
+        value == HammerCLI::NIL ? nil : value
       end
     end
   end
--- a/lib/hammer_cli/options/option_definition.rb
+++ b/lib/hammer_cli/options/option_definition.rb
@@ -23,18 +23,14 @@ module HammerCLI
       attr_accessor :deprecated_switches
 
       def initialize(switches, type, description, options = {})
-        self.value_formatter = options.delete(:format)
+        self.value_formatter = HammerCLI::Options::Normalizers::NilDecorator.new(options.delete(:format))
         self.context_target = options.delete(:context_target)
         self.deprecated_switches = options.delete(:deprecated)
         super
       end
 
       def complete(value)
-        if value_formatter.nil?
-          []
-        else
-          value_formatter.complete(value)
-        end
+        value_formatter.complete(value)
       end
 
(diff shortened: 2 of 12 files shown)


mbacovsky:
Added tests for option with defaults defined.

mbacovsky:
@akofink this may have some impact on Katello plugin, could you please take a look?

lib/hammer_cli/options/option_collector.rb
```diff
@@ -8,14 +8,28 @@ def initialize(recognised_options, option_sources)
...
         end
       end
 
+      def all_options
+        @all_options ||= translate_nils(all_options_raw)
+      end
+
       def options
-        @options ||= all_options.reject {|key, value| value.nil? }
+        @options ||= all_options.reject {|key, value| value.nil? && all_options_raw[key].nil? }
```
adamruzicka: Nitpick: Missing space between `{` and `|`

lib/hammer_cli/options/option_definition.rb
```diff
@@ -104,20 +100,23 @@ def value_description
...
+        if flag?
           Clamp.method(:truthy?)
+        else
+          formatter = value_formatter.method(:format)
+          lambda do |value|
+            if value == (ENV['HAMMER_NIL'] || HammerCLI::Options::NIL_SUBST)
+              HammerCLI::NilValue
+            else
+              formatter.call(value)
+            end
```
adamruzicka: How about turning the lambda into `OptionDefinition`'s method? It seems a bit cleaner and could be more easily tested this way.

```ruby
def default_conversion_block
  if flag?
    Clamp.method(:truthy?)
  else
    self.method(:format_or_nil)
  end
end

def format_or_nil(value)
  if value == (ENV['HAMMER_NIL'] || HammerCLI::Options::NIL_SUBST)
    HammerCLI::NilValue
  else
    value_formatter.format(value)
  end
end
```

lib/hammer_cli/options/option_definition.rb
```diff
@@ -104,20 +100,23 @@ def value_description
...
       def default_conversion_block
-        if !value_formatter.nil?
-          value_formatter.method(:format)
-        elsif flag?
+        if flag?
           Clamp.method(:truthy?)
+        else
+          formatter = value_formatter.method(:format)
+          lambda do |value|
+            if value == (ENV['HAMMER_NIL'] || HammerCLI::Options::NIL_SUBST)
```
ares: should we consider empty string as nil value? I think users would tend to do `export HAMMER_NIL=some` and they would need to do `unset` but mistake such as `HAMMER_NIL= hammer user list` would mark empty strings as nil?

lib/hammer_cli/options/option_definition.rb
ares: in the example above I suppose the formatter should be format_or_nil's argument then it makes sense to me, just the name should be improved :-)

adamruzicka: I'm afraid the result of `default_conversion_block` is somehow used by clamp and expects only one argument which is the value. But maybe I'm wrong

akofink:
@mbacovsky Thanks for pinging me. This PR doesn't appear to introduce regressions to the hammer-cli-katello plugin.

mbacovsky:
@ares, @adamruzicka updated (in separate commit)

lib/hammer_cli/options/option_definition.rb
```diff
@@ -103,20 +103,27 @@ def default_conversion_block
...
+        if value == nil_subst
+          HammerCLI::NilValue
+        else
+          value_formatter.format(value)
+        end
+      end
+
+      def nil_subst
+        nil_subst = ENV['HAMMER_NIL'] || HammerCLI::Options::NIL_SUBST
+        raise RuntimeError.new(_('Environment variable HAMMER_NIL can not be empty')) if nil_subst.empty?
```
adamruzicka: This could be simplified to just
```ruby
raise _('Environment variable HAMMER_NIL can not be empty') if nil_subst.empty?
```

mbacovsky: Thanks, fixed

mbacovsky:
@ares, @adamruzicka final touches, docs added, comments addressed.

Works as expected with the `defaults`. It is not possible to unset id parameter using its name (e.g. `--domain NIL`) but it is possible with its id option (e.g. `--domain-id NIL`). This shouldn't block this addition and needs to be addressed in hammer-cli-foreman (WIP currently).

test/functional/nil_values_test.rb
```diff
@@ -0,0 +1,75 @@
...
+  
+  it 'does not interpret NIL value when the subst is redefined' do
+    ENV.stubs(:[]).with('HAMMER_NIL').returns('NULL')
+    result = run_cmd(['--description=NIL'], {}, TestNilableCommand)
+    assert_key_val('present?', true, result)
+    assert_key_val('nil?', false, result)
+    assert_key_val('value', 'NIL', result)
+  end
+  
+  it 'overrides defaults' do
```
adamruzicka: Nitpick: this whole block uses 4-space indentation.

mbacovsky:
@adamruzicka, updated.

mbacovsky:
Thanks for the review @adamruzicka! :icecream:

PR #261, opened 2017-11-20
fixes #21707 - matches user prefix in output messages for CA certs

cfouant:


tstrachota:
Makes sense, thanks @cfouant !

PR #258, opened 2017-11-07
Fixes #7451 - Review whitespace in extracted strings

ofedoren:

PR #262, opened 2017-12-07
Revert "Fixes #7451 - Review whitespace in extracted strings"

mbacovsky:
Reverts theforeman/hammer-cli#258
I'm reverting this PR as according to the issue description there should not be the "dots" in the descriptions: 
"If I may add this too, it would also be great to review punctuation at the end of strings as this is also very inconsistent. If you type "hammer" with hammer-cli-foreman installed, many of the command descriptions have a trailing "." while others don't. I don't think it should be there."
This change also broke some tests in hammer-cli-foreman and I'd like to have PR fixing it prior the merge.

### DIFF
--- a/doc/i18n.md
+++ b/doc/i18n.md
@@ -118,14 +118,5 @@ puts _("TODO:") +
 ```
 
 
-4) It's recommended to use punctuation at the end of sentences.
-```ruby
-# CORRECT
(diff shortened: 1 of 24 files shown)


tstrachota:
Correct. I'm sorry for being too fast with the merge button.

PR #263, opened 2017-12-14
Fixes #14725 - Verbose mode does not work and is inconsistent

ofedoren:
related to https://github.com/theforeman/hammer-cli-foreman/pull/344

### DIFF
--- a/lib/hammer_cli/output/adapter/abstract.rb
+++ b/lib/hammer_cli/output/adapter/abstract.rb
@@ -7,6 +7,7 @@ module HammerCLI::Output::Adapter
     end
 
     def initialize(context={}, formatters={})
+      context[:verbosity] ||= HammerCLI::V_VERBOSE
       @context = context
       @formatters = HammerCLI::Output::Formatters::FormatterLibrary.new(filter_formatters(formatters))
       @paginate_by_default = true
--- a/lib/hammer_cli/output/output.rb
+++ b/lib/hammer_cli/output/output.rb
@@ -8,29 +8,30 @@ module HammerCLI::Output
   class Output
 
     def initialize(context={}, options={})
+      context[:verbosity] ||= HammerCLI::V_VERBOSE
       self.context = context
       self.default_adapter = options[:default_adapter]
     end
 
     attr_accessor :default_adapter
 
-    def print_message(msg, msg_params={})
-      adapter.print_message(msg.to_s, msg_params)
+    def print_message(msg, msg_params = {}, options = {})
+      adapter.print_message(msg.to_s, msg_params) if appropriate_verbosity?(:message, options)
     end
 
-    def print_error(msg, details=nil, msg_params={})
-      adapter.print_error(msg.to_s, details, msg_params)
+    def print_error(msg, details=nil, msg_params = {}, options = {})
+      adapter.print_error(msg.to_s, details, msg_params) if appropriate_verbosity?(:error, options)
     end
 
     def print_record(definition, record)
-      adapter.print_record(definition.fields, record)
+      adapter.print_record(definition.fields, record) if appropriate_verbosity?(:record)
     end
 
     def print_collection(definition, collection)
       unless collection.class <= HammerCLI::Output::RecordCollection
         collection = HammerCLI::Output::RecordCollection.new([collection].flatten(1))
       end
-      adapter.print_collection(definition.fields, collection)
+      adapter.print_collection(definition.fields, collection) if appropriate_verbosity?(:collection)
     end
 
     def adapter
@@ -64,6 +65,21 @@ module HammerCLI::Output
         formatter_list << formatter
         formatters[type] = formatter_list
       end
+    end
+
+    protected
+
+    def appropriate_verbosity?(msg_type, options = {})
+      default = case msg_type
+                when :message
+                  HammerCLI::V_VERBOSE
+                when :error
+                  HammerCLI::V_QUIET
+                when :record, :collection
+                  HammerCLI::V_UNIX
+                end
+      msg_verbosity = options[:verbosity] || default
+      context[:verbosity] >= msg_verbosity
     end
 
     private
--- a/bin/hammer
+++ b/bin/hammer
@@ -12,7 +12,8 @@ require 'hammer_cli/options/normalizers'
 # Create fake command instance to use some global args before we start.
 # Option descriptions are never displayed and thus do not require translation.
 class PreParser < Clamp::Command
-  option ["-v", "--verbose"], :flag, "be verbose"
+  option ['-v', '--[no-]verbose'], :flag, _('Be verbose (or not). True by default')
+  option ['-q', '--quiet'], :flag, _('Completely silent')
   option ["-d", "--debug"], :flag, "show debugging output"
   option ["-c", "--config"], "CFG_FILE", "path to custom config file" do |path|
     File.expand_path path
@@ -80,7 +81,7 @@ HammerCLI::Settings.load({
     :password => preparser.password,
     :host => preparser.server,
     :interactive => preparser.interactive,
-    :verbose => preparser.verbose? || preparser.debug?,
+    :debug => preparser.debug?,
     :no_headers => preparser.no_headers?,
     :reload_cache => preparser.reload_cache?,
     :verify_ssl => preparser.verify_ssl,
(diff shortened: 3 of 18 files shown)


lib/hammer_cli/output/fields.rb
```diff
@@ -19,6 +19,7 @@ def hide_blank?
     end
 
     def display?(value)
+      return false if HammerCLI.context[:verbosity].to_i < self.parameters[:verbosity].to_i
```
tstrachota: This chunk of code is repeated at multiple places. Could you please extract it into a method and move it to the parent class?

lib/hammer_cli/output/output.rb
```diff
@@ -9,11 +9,11 @@ def initialize(context={}, options={})
     attr_accessor :default_adapter
 
     def print_message(msg, msg_params={})
-      adapter.print_message(msg.to_s, msg_params)
+      adapter.print_message(msg.to_s, msg_params) if @context[:verbosity].to_i > 0
```
tstrachota: I think that we should enable specifying the verbosity per message (i.e. pass the verbosity as an argument instead of hardcoding 0).

test/functional/nil_values_test.rb
```diff
@@ -14,35 +14,35 @@ def execute
...
-  
+
   def assert_key_val(key, value, actual_output)
     assert_cmd(success_result(FieldMatcher.new(key, value.to_s)), actual_output)
   end
-  
+
   it 'accepts empty string' do
-    result = run_cmd(['--description=""'], {}, TestNilableCommand)
+    result = run_cmd(['--description=""'], { :verbosity => 1 }, TestNilableCommand)
```
tstrachota: Could you please provide some default verbosity? It's less error prone + you get rid of passing it everywhere in the tests.

tstrachota:
Thanks @ofedoren I didn't do any deep review, but I added some comments that I found at a quick glance.

tstrachota:
[test]

lzap:
There is defacto standard in UNIX which says that if a program successfully does its job it writes nothing to the stdout or stderr so it's by default "scriptable". Then usually there is a `--verbose` flag or similar to increase verbosity and more levels are added via repeating the same option over and over again, e.g. `-vvvv` usually means very verbose, but each program implements different options.

What you have is probably fine, I tend to prefer `--verbose / -v`, `--debug / -d` and `--trace` options as the three good to go options. But not a strong opinions, this depends. Whatever we do, I vote for *zero output* by default.

mbacovsky:
@ofedoren, @tstrachota  first of all, I really appreciate your effort on this, as this is uneasy and aged debt that needs to be solved. I've played a bit with hammer with this PR, so let me share my findings.

The logic still seems to be somehow complex especially in context of existing switches. I tried to put together my requirements, to what I expect,  hopefully in line what @lzap requested in the original issue. So "As an user I want"
-  **hammer output to be completely silent**
   - here I care only about the exit code or don't care about the result at all
   - does the use case differ from ` --output silent`? Do we need both?
   - `--verbosity 0`
- **hammer output to contain exactly what I asked for**
   - linux natural mode
   - no output for modifications
   - only data I asked for for `list` and `show`
   - errors to STDERR
   - `--verbosity 1`
- **hammer output to contain verbose data improving my comfort and confidence**
   - confirmation in modifications 
   - pagination information
   - further details (do we have any examples?)
   - errors to STDERR
   - `--verbosity 2`
   - `--verbose`, `-v`
   - current behaviour
   - this should stay the default to not change the behavior
- **to be able to set my default verbosity in configuration**
   - this way I can set `--verbosity 1` to get what @lzap requested and use `-v` to get more verbose output
- **hammer to overwhelm me with information for debugging**
   - `--debug`, `-d`
   - `--verbosity 2` + `--log-level DEBUG` + log > STDOUT
   - not sure if this qualify for verbosity 3 as it is rether not more granular output but mixture of two sources. My preference is not to make it level 3.
   - log to STDOUT makes it possible to discard or redirect STDERR to file to not see duplicite errors (STDERR and log)

Reasonable user settings would be `verbosity = 1` in hammer config then `-v` and `-d` would work as expected from a traditional linux CLI user.

From what I defined above I'm not sure if it is reasonable to mix verbosity level with columns being displayed. It seems orthogonal to the verbosity settings as I may need all the columns displayed along with `--verbosity 1`. It is also difficult to define which columns should be what verbosity level and may differ from user to user. I'd suggest to address this feature independently in  https://github.com/theforeman/hammer-cli/pull/276.

mbacovsky:
I forgot to add
 - **As a user I want to see `-h|--help` output regardless on verbosity level set**
   - might make sense to show some warning that help is displayed even though the verbosity is set to 0

The current PR defaults to 0 due to a issue @tstrachota pointed out. It took me while I figured out. Why I  see no output

tstrachota:
@mbacovsky I came to the same conclusion when I was thinking about it. We've been talking about it with @ofedoren offline.

I think that it's two separate things (and two separate options):
1. verbosity of the messages
2. verbosity of the data being printed

The first one should be set via the `--verbosity` parameter. The latter should imho have it's own parameter, e.g. `--detail` seems good name to me. Even though users can always ask exactly for options that the want to see, `--detail` would offer them reasonable presets. For example:
- `--detail=thin` - only id and name (can be rendered faster if API supports `thin`)
- `--detail=default` - standard setup
- `--detail=full` - everything

So, from the users' perspective it behaves as a separate feature, but the implementation is the same. We only need to add one more context variable that would be switching the details.

mbacovsky:
@ofedoren, @tstrachota, Let me sum up result of our offline discussion. Feel free to correct me and add missing items.

- verbosity level is internal thing, `--verbosity` shall not be exposed to users
- verbosity 0
  - nothing to stdout just exitcode and errors to STDERR
  - `--silent` to switch level 0. `-s` is already taken, so we may consider `--quiet, -q`
  - probably does not make sense to have option for this in config
  - `--help` prints even in --silent mode (makes sense ?)
- verbosity 1
  - no output for modifications
  - just data and print_message with :verbosity => 1, by default verbosity of print_message is 2 (?)
  - interactive I/O (passwd prompt etc) should be seen
  - no progress bars in async tasks [BZ#1563129](https://bugzilla.redhat.com/show_bug.cgi?id=1563129)
  - `--no-verbose` or `verbose = false` in config
- verbosity 2
  - the default and current behavior
  - `--verbose` to override `verbose=false`
  - data + print_messages up to :verbosity=>2, interactive I/O, progress bars
  - errors to STDERR
- there is room for future verbosity 3 if needed
  - --very-verbose
  - verbosity 2 + print_massage up to :verbosity=>3
- fields are not verbosity aware (they print or not as a whole), filtering of fields shall be solved separately
- document verbosity levels and related options in the project
- create issue for addition to Foreman manual

tstrachota:
I agree with the verbosity levels as you described them @mbacovsky.

I'd avoid having a separate switch for each verbosity level in the config files. Especially if we decide to add "very verbose" mode in future, it would result in having:
```yaml
verbose: false
very_verbose: true
```

I'd prefer having a single config item with multiple values, for example:
```yaml
# Verbosity of the commands. Possible values:
#   quiet - ...
#   default - ...
#   verbose - ...
verbosity: default
```

ofedoren:
updated.

Interactive I/O will be shown even with --quiet option in a case when credentials are not specified in the config file.
Also I've prepared a [PR](https://github.com/theforeman/hammer-cli-foreman-tasks/pull/36) to [hammer-cli-foreman-tasks](https://github.com/theforeman/hammer-cli-foreman-tasks) which hides progress bar.

lib/hammer_cli/output/adapter/abstract.rb
```diff
@@ -7,6 +7,7 @@ def tags
     end
 
     def initialize(context={}, formatters={})
+      context[:verbosity] ||= HammerCLI::V_VERBOSE
```
tstrachota: I'm not sure I like setting default verbosity into context here. It should ideally be set only at one place. @mbacovsky what's your take on this?

lib/hammer_cli/output/output.rb
```diff
@@ -8,29 +8,30 @@ def to_s
   class Output
 
     def initialize(context={}, options={})
+      context[:verbosity] ||= HammerCLI::V_VERBOSE
```
tstrachota: Ditto setting default at multiple places.

tstrachota:
@mbacovsky please take a look if the changes are in line with what we talked about.
From my perspective it's correct. Only some minor tuning and some more tests are needed.

@ofedoren++

ofedoren:
@tstrachota added some tests.

tstrachota:
@ofedoren thanks for adding the tests, rebase is needed.
@mbacovsky ping, can you please confirm the changes are ok from your point of view?

ofedoren:
@tstrachota, [done](https://github.com/theforeman/hammer-cli/blob/3725e036b020457112f52d33def83c0ebe1fc9bb/doc/creating_commands.md#verbosity). Please, let me know if I forgot to mention something.

tstrachota:
@mbacovsky ?

tstrachota:
This has been approved for more than two weeks. I suppose that's been enough time for everybody to raise their concerns. I'm going to merge the PR now.

Thanks @ofedoren! That was a lot of work. Please consider this feature for the next community demo.

PR #264, opened 2017-12-18
Refs #21768 - Make option sources overridable

mbacovsky:


mbacovsky:
@tstrachota, updated. I tried to record my findings about option evaluation in the docs. Does it makes sense? Anything important worth adding?

tstrachota:
:+1: great, thank you for the docs!

PR #266, opened 2018-01-26
Refs #7451 - tests for message formats

tstrachota:
Tests that check that:
- option descriptions **don't end** with '.'
- command descriptions **don't end** with '.'
- success messages in commands **end** with '.'
- failure messages in commands **don't end** with '.' (because they are glued with ":" and details)

### DIFF
--- a/test/unit/messages_test.rb
+++ b/test/unit/messages_test.rb
@@ -1,0 +1,7 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+require 'hammer_cli/testing/messages'
+
+include HammerCLI::Testing::Messages
+describe "message formatting" do
+  check_all_command_messages(HammerCLI::MainCommand)
+end
--- a/lib/hammer_cli/testing/messages.rb
+++ b/lib/hammer_cli/testing/messages.rb
@@ -1,0 +1,54 @@
+module HammerCLI
+  module Testing
+    module Messages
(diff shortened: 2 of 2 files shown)


tstrachota:
I'm not sure what was the final agreement about success messages? Should they contain periods or not? @mbacovsky, @akofink ?

akofink:
I think so. They're technically very short sentences (i.e.):

Organization created.
User updated.

tstrachota:
I also lean towards having periods in the success messages. Agreed then.

test/unit/messages_test.rb
```diff
@@ -0,0 +1,7 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+require 'hammer_cli/testing/messages'
+
+include HammerCLI::Testing::Messages
+describe "message formatting" do
+  check_commands(HammerCLI::MainCommand)
```
mbacovsky: @tstrachota while we are including it in a global space I'd prefer more specific  name. Also for test readability it should suggest what we are testing. How about `check_command_messages`?

tstrachota:
Makes sense @mbacovsky, updated.

tstrachota:
@mbacovsky ping

tstrachota:
[test]

mbacovsky:
@tstrachota, thanks for the rename.  LGTM :+1:  :shark:

PR #267, opened 2018-02-09
Remove legacy Gemfile code for Ruby < 2.0

d-m-u:
@tstrachota Hi! This is for Refactor #21359 (http://projects.theforeman.org/issues/21360). I couldn't find the other file that's linked in that issue.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,11 +3,6 @@ source "https://rubygems.org"
 gemspec
 
 gem 'gettext', '>= 3.1.3', '< 4.0.0'
-if RUBY_VERSION < '2.1.0'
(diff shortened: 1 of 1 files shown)


tstrachota:
Thanks for your contribution @d-m-u 👍
In fact, the issue for `hammer-cli` part of the problem is http://projects.theforeman.org/issues/21359 You referenced `hammer-cli-foreman` one, that's why you couldn't find the other code.

I split the issue into two and merged your PR as: https://github.com/theforeman/hammer-cli/commit/651bc98c3af7494421d5214e283ff13d93e0e4d4

PR #269, opened 2018-02-12
Fixes #21359 - Removes code for ruby < 2.1.0

d-m-u:
Hammer no longer supports ruby 2.0, there's no point keeping the old code.

### DIFF
--- a/lib/hammer_cli/output/adapter/csv.rb
+++ b/lib/hammer_cli/output/adapter/csv.rb
@@ -1,13 +1,3 @@
-require 'csv'
-if CSV.const_defined? :Reader
-  # Ruby 1.8 compatible
-  require 'fastercsv'
-  Object.send(:remove_const, :CSV)
-  CSV = FasterCSV
-else
-  # CSV is now FasterCSV in ruby 1.9
-end
-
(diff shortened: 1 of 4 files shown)


d-m-u:
@tstrachota thanks! This includes the removal of a method in /test/unit/i18n_test.rb that was for testing the old fast_text gem that I think should also be removed. Lemme know if I'm wrong and I'd be happy to fix!

tstrachota:
[test]

tstrachota:
Thanks for fixing the rest, @d-m-u !
We'll definitely need to remove the tests for `String#format` with this PR: https://github.com/theforeman/hammer-cli/blob/master/test/unit/utils_test.rb#L13 I'm also afraid it might still be used at some places. Would you mind checking it?

mmoll:
What's the status here?

tstrachota:
The PR had wrong label. It's waiting for contributor. Tests are failing and need to be updated as I mentioned in my previous comment (@d-m-u ping).

mbacovsky:
Closing this PR as it was already merged as part of https://github.com/theforeman/hammer-cli/pull/285.

PR #270, opened 2018-02-12
Refs #22554 - Pin clamp version to < 1.2.0

mbacovsky:
There are some tests failing due to updated clamp version.
Temporarily pinning the version to unblock the tests.


tstrachota:
👍 thanks @mbacovsky

PR #271, opened 2018-02-22
Fixes #21514 - Add personal access token cli support

ofedoren:
Needs [Main PR](https://github.com/theforeman/hammer-cli-foreman/pull/350)  to be merged at first.


tstrachota:
Thanks @ofedoren since this filter is `hammer-cli-foreman` specific, it should be moved to it's repo and merged together with the rest of the access token related code.

tstrachota:
Thanks @ofedoren since this filter is `hammer-cli-foreman` specific, it should be moved to it's repo and merged together with the rest of the access token related code.

PR #272, opened 2018-03-02
Fixes #22752 - correct parsing of enum types

tstrachota:


tstrachota:
@mbacovsky can you take a look, please?

mbacovsky:
@tstrachota, the code looks good. Thanks for cleaning the logic.

While testing I was not able to find the reproducer and it is not clear from the issue either. I assume this was spotted in smart class parameters while adding override value. The documentation is now fixed to contain "Must be one of types: ..." which does not match anyway.
I tested the add-matcher's value param is parsed properly and that the other enums stay unaffected - all passed nice. Merging.

tstrachota:
@mbacovsky thanks. There's no reproducer in the Foreman afaik now. It has been fixed.
The second case in the issue is from Katello: https://github.com/Katello/katello/pull/7220

PR #275, opened 2018-05-31
Fixes #21634 - Allows modification of output fields

ofedoren:


tstrachota:
Thanks @ofedoren! It looks good at a first glance. Would you mind adding some docs with examples how to change a commands output from a plugin? It would make testing easier plus we can refer to them in future.

ofedoren:
@tstrachota updated. I did some changes comparing to the last version of this commit and added some [docs](https://github.com/ofedoren/hammer-cli/blob/feat-21634-modif-output-fields/doc/commands_modification.md).

lib/hammer_cli/output/definition.rb
```diff
@@ -12,10 +12,117 @@ def append(fields)
...
+      return deep_delete!(@fields, by) if by[:deep_delete]
+      ids, labels = adjust_options(by)
+      @fields.delete_if do |f|
+        label = f.label
+        label = f.label.downcase unless label.nil?
+        ids.include?(f.id) || labels.include?(label)
+      end
+    end
+
+    def select_fields!(by = {})
```
tstrachota: I'm not sure if there's some real-world use-case for this method. I can't imagine that somebody would want to keep just some fields and remove the rest.

It also doesn't work as expected. When I tested it on host info
```ruby
HammerCLIForeman::Host::InfoCommand.output_definition.select_fields!(:labels => ['Id'], :deep_select => true)
```
the info command then did contain only two "Id" fields, but none of them displayed any data.

I think it's fine to simple remove this method.

lib/hammer_cli/output/definition.rb
```diff
@@ -12,10 +12,117 @@ def append(fields)
...
+      return @fields = deep_select(@fields, by) if by[:deep_select]
+      ids, labels = adjust_options(by)
+      @fields.keep_if do |f|
+        label = f.label
+        label = f.label.downcase unless label.nil?
+        ids.include?(f.id) || labels.include?(label)
+      end
+    end
+
+    def replace_field(field_id, fields = nil, &block)
```
tstrachota: This method is handy! I like using the block.
Does it also accept a single field? Using is as `replace_field(:name, Field.new(:key => :surname))` would be nice too.

lib/hammer_cli/output/definition.rb
```diff
@@ -12,10 +12,117 @@ def append(fields)
...
+      @fields = []
+    end
+
     def empty?
       @fields.empty?
     end
 
+    private
+
+    def build_definition(fields = nil, &block)
```
tstrachota: We're using similar code at 2 places already. I think we could refactor it into a definition's public method:
```ruby
def build(dsl = Dsl.new, &block)
  dsl.build(&block) if block_given?
  append(dsl.fields)
end
```
and also if we changed append to:
```ruby
def append(fields)
  unless fields.nil?
    fields = [fields] unless fields.is_a? Array
    @fields += fields
  end
end
```
than it would be much easier for use at all the places and also decoupled from the dsl definition.

This method could then be just:
```ruby
definition = self.class.new
definition.append(fields)
definition.build(&block)
definition
```

lib/hammer_cli/output/definition.rb
```diff
@@ -12,10 +12,117 @@ def append(fields)
...
+    def replace_field(field_id, fields = nil, &block)
+      i = @fields.find_index do |f|
+        f.id == field_id
+      end
+      return if i.nil?
+      definition = build_definition(fields, &block)
+      multiple_replace(i, definition.fields)
+    end
+
+    def at(path = [])
```
tstrachota: I like that method. Could the path also accept a single symbol? Eg `at(:interfaces)`

doc/commands_modification.md
```diff
@@ -0,0 +1,82 @@
...
+
+```ruby
+# Options you might use with select/delete_fields method
+# should be a hash, where you specify which fields should be
+# selected or deleted from command's or field's output definition.
+options = {
+  deep_select: true/false,
+  deep_delete: true/false,
+  keys: [comma separated :field_key or/and :field_id],
+  labels: [comma separated 'field_label']
```
tstrachota: I guess you could write just "array"

lib/hammer_cli/output/definition.rb
```diff
@@ -12,10 +12,117 @@ def append(fields)
       @fields += fields
     end
 
+    def delete_fields!(by = {})
+      return clear if by.empty?
+      return deep_delete!(@fields, by) if by[:deep_delete]
```
tstrachota: Use just `by[:deep]` please. "delete" feels a bit redundant there.

ofedoren:
@tstrachota, updated.

If there is no need for select_fields! method, is there need for delete_fields! method then? I think they are very similar, so I removed it as well. I guess if there will be need to delete a field, it can be done via ```d.at(path).insert(:replace, field_id_to_delete) do; end```

Also now it's possible to insert before/after or replace a field via one single method (see updated [docs](https://github.com/ofedoren/hammer-cli/blob/feat-21634-modif-output-fields/doc/commands_modification.md)). I guess it's a good way to avoid code repetition, which in this case will be x3.

lib/hammer_cli/output/definition.rb
```diff
@@ -8,14 +8,72 @@ def initialize
       @fields = []
     end
 
+    def build(dsl = Dsl.new, &block)
+      dsl.build(&block) if block_given?
+      append(dsl.fields)
+    end
+
     def append(fields)
```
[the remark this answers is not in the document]
tstrachota: Note: it would also make code in `insert` simpler:
```
definition.append(fields, &block)
```

lib/hammer_cli/output/definition.rb
```diff
@@ -8,14 +8,72 @@ def initialize
...
+      return @fields if fields.nil?
+      fields = [fields] unless fields.is_a? Array
       @fields += fields
     end
 
+    def find_field(field_id)
+      @fields[field_index(field_id)]
+    end
+
+    def insert(mode = :after, field_id = nil, fields = nil, &block)
```
tstrachota: How about changing the signature to `def insert(mode, field_id, fields = nil, &block)` ?
Without `field_id` the method will fail anyway so both parameters are required.

lib/hammer_cli/output/definition.rb
```diff
@@ -8,14 +8,72 @@ def initialize
...
+      index = field_index(field_id)
+      index += 1 if mode == :after
+      definition = self.class.new
+      definition.append(fields)
+      definition.build(&block)
```
tstrachota: This seems too complicated to me. I guess the only reason why you created new definition is the fact that `build` automatically appends the fields. How about moving injection of `Dsl.new` into constructor and using just `@dsl.build` here?

ofedoren:
@tstrachota, updated. Thanks for suggestions for better method definitions!

[Updated docs](https://github.com/theforeman/hammer-cli/blob/dff778b4096831624c0c407e865ab8291ba3f61c/doc/commands_modification.md)

Also, before it gets merged, I want to be sure if this is an acceptable way of identifying a field by default:
  * https://github.com/theforeman/hammer-cli/pull/275/files#diff-5429b05e7a5cdcd85f448cc1b33defe9R16
  * https://github.com/theforeman/hammer-cli/pull/275/files#diff-cdbcef2e121565465d7438e227ec3f48R23

lib/hammer_cli/output/fields.rb
```diff
@@ -4,14 +4,16 @@ module Fields
 
   class Field
 
-    attr_reader :label
-    attr_reader :path
+    attr_accessor :label
```
mbacovsky: Do we want to enable the writers and allow change these form outside?

lib/hammer_cli/output/dsl.rb
```diff
@@ -20,6 +20,7 @@ def build(&block)
     def field(key, label, type=nil, options={}, &block)
       options[:path] = current_path.clone
       options[:path] << key if !key.nil?
+      options[:field_id] = options[:field_id] || key || options[:key]
```
mbacovsky: This is inconsistent with the Field class defaults. If field is created directly without the `:field_id` the default id is 'label'. Via dsl the 'key' wins. What about not setting the default value in here and leave it up to Field class?

lib/hammer_cli/output/fields.rb
```diff
@@ -4,14 +4,16 @@ module Fields
...
+    attr_accessor :path
+    attr_accessor :id
 
     def initialize(options={})
       @hide_blank = options[:hide_blank].nil? ? false : options[:hide_blank]
       @hide_missing = options[:hide_missing].nil? ? true : options[:hide_missing]
       @path = options[:path] || []
       @label = options[:label]
+      @id = options[:field_id] || label || options[:key]
```
mbacovsky: Is there a reason the option is called `:field_id` and the attribute just `id`? It is not consistent with other options.

lib/hammer_cli/output/fields.rb
ofedoren: Since we're [using labels](https://github.com/theforeman/hammer-cli/pull/275/files#diff-5429b05e7a5cdcd85f448cc1b33defe9R16) to set field's ID, which is used to address a field, I don't think it's a good idea now :)

But again, is [this](https://github.com/theforeman/hammer-cli/pull/275/files#diff-5429b05e7a5cdcd85f448cc1b33defe9R16) an acceptable way? We could allow label changing from outside, but then we shouldn't identify fields by labels I guess..

mbacovsky: Regarding setting the default for id I think it is okay. Maybe with 'key' preferred over the 'label'.
In the original comment I was referring more to all three attr_accessors in general. It does not feel correct to modify any of these attributes on existing field.  Wouldn't it be better to replace filed with a new field?

I can imagine changing label in container or collection could be awkward, so ability to change label might be helpful in some cases. The :id would need to be a method getting the id dynamically.

mbacovsky:
@ofedoren, I've left some comments and questions inline. Besides that it looks nice and powerful.

ofedoren:
Updated with @mbacovsky suggestions.

mbacovsky:
@ofedoren thanks for quick update!

I'm sorry I didn't noticed earlier but I played with the feature and wanted to found out how it behaves when the id is not found. Similarly as in help extension it raises error which prevents whole plugin from loading.

I'd prefer it to print warning, log whatever we have for debuging later and print as much as we can. The reason is to prevent e.g. one removed field in hammer-cli-foreman from loading whole other plugin.
It may be a bit difficult to handle the errors with the current notation. Would it help to make it similar to extend_help and wrap it into a block like: `CommandXY.extend_output_definition do |od|; end`?

ofedoren:
@mbacovsky, updated.

I added [handling](https://github.com/theforeman/hammer-cli/pull/275/files#diff-cbbc9e39db2dacf9c728b47b3a066ae0R96) similar to [one from help extension](https://github.com/theforeman/hammer-cli/pull/277/files#diff-cbbc9e39db2dacf9c728b47b3a066ae0R93). So, if a field is not found, then the whole output extension block will fail, but the plugin will be loaded and other extension blocks will be applied (if there are no errors).

Also, [updated docs](https://github.com/theforeman/hammer-cli/blob/72ad3632a40d676decf10fdc5eab1c829c8576c0/doc/commands_modification.md).

mbacovsky:
[test hammer]

mbacovsky:
@ofedoren a way better, thanks! 

While testing I found another issue. I had the `:mark_translations` settings turned on so my labels were modified to e.g. `>Name<` and I realized addressing by labels won't work for localized Hammer. But there is not much we can do about it, right?

I also found out that when the field is created via dsl it won't have the `:key` set which limits the possibilities to identify the fields. The following patch fixed it for me:
```diff
diff --git a/lib/hammer_cli/output/dsl.rb b/lib/hammer_cli/output/dsl.rb
index e0cf7ab..5cf6a38 100644
--- a/lib/hammer_cli/output/dsl.rb
+++ b/lib/hammer_cli/output/dsl.rb
@@ -19,7 +19,10 @@ module HammerCLI::Output
 
     def field(key, label, type=nil, options={}, &block)
       options[:path] = current_path.clone
-      options[:path] << key if !key.nil?
+      unless key.nil?
+        options[:path] << key
+        options[:key] = key
+      end
 
       options[:label] = label
       type ||= Fields::Field
```

Otherwise it looks good!

lib/hammer_cli/output/definition.rb
```diff
@@ -1,21 +1,68 @@
...
     def empty?
       @fields.empty?
     end
 
-  end
+    private
 
+    def field_index(field_id)
+      index = @fields.find_index do |f|
+        f.id == field_id
```
mbacovsky: I had an idea to match the field by ANY of those identifiers like:
```ruby
f.match_id?(field_id)
```
where the matcher is defined as
```ruby
def match_id?(field_id)
  @options[:id] == filed_id || @options[:key] == field_id || @label == field_id
end
```
It could prevent breaking the extension when e.g. the field_id is added.
What do you think?

ofedoren:
@tstrachota, updated.

lib/hammer_cli/i18n.rb
```diff
@@ -146,4 +146,3 @@ def self.main_makefile
 
 domain = [HammerCLI::I18n::LocaleDomain.new, HammerCLI::I18n::SystemLocaleDomain.new].find { |d| d.available? }
 HammerCLI::I18n.add_domain(domain) if domain
-
```
tstrachota: Please avoid changing files that aren't affected by the purpose of the patch.

tstrachota:
[test hammer]

tstrachota:
@mbacovsky ping

tstrachota:
[test hammer]

tstrachota:
All comments have been addressed and the changes work fine. Merging, thanks @ofedoren this is an important feature!

mbacovsky:
Thanks @ofedoren, @tstrachota!

PR #276, opened 2018-06-08
Fixes #19135 - Possibility to limit fields that are displayed

ofedoren:
Requires #275 to be merged first. PR already goes with that commit for testing.

The [issue](http://projects.theforeman.org/issues/19135) suggests --columns option while I suggest --only, because it works with all the adapters (not for table nor list commands only) and it's shorter :)

Also it accepts labels, which user sees, as arguments, so as mentioned in the issue e.g.:
`hammer subnet list --columns name,vlanid` will not work, while
`hammer --only name,"vlan id" subnet list` will work.

### DIFF
--- a/doc/creating_commands.md
+++ b/doc/creating_commands.md
@@ -164,6 +164,14 @@ In cases when you want to deprecate just one of more possible switches use the e
     :deprecated => { '--name' => _('Use --alias instead') }
 ```
 
+#### Predefined options
+Also Hammer offers predefined options now. Those are just options, but with
+predefined functionality. To define them in your command use
+`use_option :option_name` method.
+
+Here is the list of predefined options:
+  * `:fields` Expects a list with fields to show in output, see [example](creating_commands.md#printing-hash-records).
+
 
 ### Option builders
 Hammer commands offer option builders that can be used for automatic option generation.
@@ -523,8 +531,12 @@ Imagine there's an API of some service that returns list of users:
 ```
 
 We can create an output definition that selects and formats some of the fields:
+
+_NOTE_: Every field can be arranged in so-called field sets. All the fields by default go to `'DEFAULT'` and `'ALL'` sets. Fields which are in the `'DEFAULT'` set will be printed by default. To see printed other field sets, use predefined option `--fields NAME`, where `NAME` is a field set name in ALLCAPS.
 ```ruby
 class Command < HammerCLI::AbstractCommand
+  # To be able to select fields which should be printed
+  use_option :fields
 
   output do
     # Simple field with a label. The first parameter is the key in the printed hash.
@@ -536,7 +548,7 @@ class Command < HammerCLI::AbstractCommand
     field :roles, 'System Roles', Fields::List
 
     # Label is used for grouping fields.
-    label 'Contacts' do
+    label 'Contacts', sets: ['ADDITIONAL', 'ALL'] do
       field :email, 'Email'
       field :phone, 'Phone No.'
     end
@@ -565,18 +577,38 @@ Using the base adapter the output will look like:
 ID:            1
 System Roles:  Admin, Editor
 Name:          Tom Sawyer
-Contacts:
-  Email:       tom@email.com
-  Phone No.:   123456111
 Created At:    2012/12/18 15:24:42
 
 ID:            2
 System Roles:  Admin
 Name:          Huckleberry Finn
+Created At:    2012/12/18 15:25:00
+```
+
+Using the base adapter with `--fields ALL` or `--fields DEFAULT,ADDITIONAL` the output will look like:
+```
+ID:            1
+System Roles:  Admin, Editor
+Name:          Tom Sawyer
+Created At:    2012/12/18 15:24:42
+Contacts:
+  Email:       tom@email.com
+  Phone No.:   123456111
+
+ID:            2
+System Roles:  Admin
+Name:          Huckleberry Finn
+Created At:    2012/12/18 15:25:00
 Contacts:
   Email:       huckleberry@email.com
   Phone No.:   123456222
-Created At:    2012/12/18 15:25:00
+```
+
+_NOTE_: `--fields` as well lets you to print desired fields only. E.g. to see the users' emails without any additional information use `--fields contacts/email`:
+```
+Email:       tom@email.com
+
+Email:       huckleberry@email.com
 ```
 
 You can optionally use the output definition from another command as a base and extend it with
--- a/lib/hammer_cli/output/adapter/abstract.rb
+++ b/lib/hammer_cli/output/adapter/abstract.rb
@@ -6,11 +6,12 @@ module HammerCLI::Output::Adapter
       []
     end
 
-    def initialize(context={}, formatters={})
+    def initialize(context={}, formatters={}, filters = {})
       context[:verbosity] ||= HammerCLI::V_VERBOSE
       @context = context
       @formatters = HammerCLI::Output::Formatters::FormatterLibrary.new(filter_formatters(formatters))
       @paginate_by_default = true
+      @filters = filters
     end
 
     def paginate_by_default?
@@ -41,10 +42,14 @@ module HammerCLI::Output::Adapter
       raise NotImplementedError
     end
 
+    def reset_context
+      @context.delete(:fields)
+    end
+
     protected
 
-    def field_filter
-      HammerCLI::Output::FieldFilter.new
+    def filter_fields(fields)
+      HammerCLI::Output::FieldFilter.new(fields, field_filters)
     end
 
     def self.data_for_field(field, record)
@@ -71,15 +76,23 @@ module HammerCLI::Output::Adapter
       $stdout
     end
 
-    def displayable_fields(fields, record, compact_only: false)
-      fields.select do |field|
-        field_data = data_for_field(field, record)
-        if compact_only && !field_data.is_a?(HammerCLI::Output::DataMissing)
-          true
-        else
-          field.display?(field_data)
-        end
+    def field_filters
+      {
+        classes_filter: classes_filter,
+        sets_filter: sets_filter
+      }.merge(@filters) do |_, old_filter, new_filter|
+        old_filter + new_filter
       end
+    end
+
+    def classes_filter
+      return [] if @context[:show_ids]
+
+      [Fields::Id]
+    end
+
+    def sets_filter
+      @context[:fields] || []
     end
 
     private
@@ -94,6 +107,5 @@ module HammerCLI::Output::Adapter
         map
       end
     end
-
   end
 end
--- a/lib/hammer_cli/output/adapter/tree_structure.rb
+++ b/lib/hammer_cli/output/adapter/tree_structure.rb
@@ -1,7 +1,6 @@
 module HammerCLI::Output::Adapter
   class TreeStructure < Abstract
-
-    def initialize(context={}, formatters={})
+    def initialize(context = {}, formatters = {}, filters = {})
       super
       @paginate_by_default = false
     end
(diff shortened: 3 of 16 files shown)


lib/hammer_cli/main.rb
```diff
@@ -39,6 +39,9 @@ class MainCommand < AbstractCommand
     option ["--interactive"], "INTERACTIVE", _("Explicitly turn interactive mode on/off"),
       :format => HammerCLI::Options::Normalizers::Bool.new,
       :context_target => :interactive
+    option ["--only"], "FIELDS", _("Show selected columns or fields only"),
```
tstrachota: Placing the option in the main command isn't very intuitive. Could you move it to the base info and list commands? It also doesn't make sense to use it for create, update and delete.

lib/hammer_cli/output/output.rb
```diff
@@ -69,5 +71,11 @@ def init_adapter(adapter_name)
       @adapter ||= self.class.adapters[adapter_name].new(context, self.class.formatters)
     end
 
+    def adjust_definition(definition)
+      return unless @context.key?(:only)
+      return if adapter.class <= HammerCLI::Output::Adapter::TreeStructure
+      labels = @context[:only].map(&:downcase)
+      definition.select_fields!(deep_select: true, labels: labels)
```
tstrachota: Using `select_fields!` will remove the fields from the command class for good. This will disallow using the option repeatedly in shell mode:
```
hammer shell
> host list --fields Id,Name
... will list the hosts
> host list --fields Id,Name,Subnet
... will break
```
Note that at the moment it's not possible to use the option in shell mode at all (see my previous comment).

lib/hammer_cli/output/output.rb
```diff
@@ -69,5 +71,11 @@ def init_adapter(adapter_name)
       @adapter ||= self.class.adapters[adapter_name].new(context, self.class.formatters)
     end
 
+    def adjust_definition(definition)
+      return unless @context.key?(:only)
+      return if adapter.class <= HammerCLI::Output::Adapter::TreeStructure
```
tstrachota: Ideally we should deal with filtering options at one place only.

ofedoren:
updated.

Field sets are defined as new output definitions for commands. Field sets could be defined with `output(nil || output_definition, 'SET_NAME') do ... end`. Field set 'ALL' is a special one which is defined by default as one field set created from all currently available in `Command.field_sets`. All the fields currently defined with `output do ... end` go to the 'DEFAULT' field set.

for testing wth HammerCLIForeman:

Now it requires something like:
```
use_option :fields

output(nil, 'THIN') do
  field :id, _('Id')
  field :name, _('Name')
end
```

in commands such as `HammerCLIForeman::ListCommand` and `HammerCLIForeman::InfoCommand` to get the feature work. Also to enable 'thin' request we need to include
`      params_pruned['thin'] = true if current_output_definition.thin?`
in [request_params](https://github.com/theforeman/hammer-cli-foreman/blob/a6f940b32e77b863c678163e3c230c11737ef88e/lib/hammer_cli_foreman/commands.rb#L181)

I've prepared a PR with those changes for HammerCLIForeman. If there is a need, I can send it.

tstrachota:
Thank you @ofedoren. I'm a bit concerned about the design. Doesn't adding multiple output definitions go against your PR for output modification? https://github.com/theforeman/hammer-cli/pull/275

Wouldn't it be easier if each field would have notion about which field sets it should be printed in? It should then be just matter of filtering the fields before they're printed and still only one output definition would be used.

ofedoren:
@tstrachota, you're right, it was unnecessary complicated, so I redesignied according to your suggestions:
 - Field sets are specified in options: `field :key, :label, :type, sets: ['SETNAME']`
 - Only one output definition is used
 - Fields filter does the whole work

[Updated docs](https://github.com/theforeman/hammer-cli/blob/1884b36c6f1bfdb3b894de4b623e9d8d28238849/doc/creating_commands.md)

Still requires `params_pruned['thin'] = true if output.adapter.thin_context?`
in [request_params](https://github.com/theforeman/hammer-cli-foreman/blob/a6f940b32e77b863c678163e3c230c11737ef88e/lib/hammer_cli_foreman/commands.rb#L181) for thin request and `use_option :fields` in `ListCommand`s and `InfoCommand`s 

Also, I'm afraid we need to specify `'THIN'` set explicitly for `Id` and `Name` fields. I was thinking about adding `Id` and `Name` fields to `THIN` set by default from core just looking if field name is `id` or `name`, but I don't think it's a good idea.

tstrachota:
This needs a rebase.

ofedoren:
@tstrachota, rebased.

doc/creating_commands.md
```diff
@@ -100,11 +100,14 @@ Option declaration is the same as in clamp so please read it's
...
 conflict with methods of the commands. So to access value of an `--name` option you have to call `option_name()`
 
+Also Hammer offers predefined options now. To define them in your command use `use_option :option_name` method.
+[Here](https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/options/predefined.rb#L7) is the list of predefined options.
 
 Example option usage could go like this:
 ```ruby
 class HelloCommand < HammerCLI::AbstractCommand
 
+  use_option :option_name, :or_names
```
mbacovsky: @ofedoren Here I'd expect naming conflict with the option as both could be called `--name`, also from the example output it is not apparent what it does. There is just one --name option in the output. Maybe it would be better not to mention `use_option` here and create separate subsection  `####Predefined options` next to `#### Deprecated options`. May make sense to explain how it works, how to use it and list available predefined options. With `:fields` we can add brief description with reference to `Printing hash records` section for more details.

lib/hammer_cli/options/option_definition.rb
```diff
@@ -27,9 +27,9 @@ class OptionDefinition < Clamp::Option::Definition
       attr_accessor :deprecated_switches
 
       def initialize(switches, type, description, options = {})
-        self.value_formatter = options.delete(:format) || HammerCLI::Options::Normalizers::Default.new
-        self.context_target = options.delete(:context_target)
-        self.deprecated_switches = options.delete(:deprecated)
+        self.value_formatter = options[:format] || HammerCLI::Options::Normalizers::Default.new
```
mbacovsky: Is there a reason to remove the cleanup of the options?

lib/hammer_cli/output/adapter/abstract.rb
```diff
@@ -41,10 +42,19 @@ def print_collection(fields, collection)
       raise NotImplementedError
     end
 
+    def thin_context?
+      return false if @context[:fields].nil?
```
mbacovsky: Wouldn't `@context[:fields] ==['THIN']` do the same as the two lines here?

lib/hammer_cli/output/adapter/tree_structure.rb
```diff
@@ -34,15 +34,11 @@ def prepare_message(msg, msg_params = {})
...
-    def field_filter
-      filtered = []
-      filtered << Fields::Id unless @context[:show_ids]
-      HammerCLI::Output::FieldFilter.new(filtered)
-    end
-
     def render_fields(fields, data)
-      fields = field_filter.filter(fields)
-      fields = displayable_fields(fields, data)
+      fields = filter_fields(fields).filter_by_classes
```
mbacovsky: Calling the filters repeats in all the adapters, could it be DRY'ed out a bit?

lib/hammer_cli/output/output.rb
```diff
@@ -25,13 +25,15 @@ def print_error(msg, details=nil, msg_params = {}, options = {})
 
     def print_record(definition, record)
       adapter.print_record(definition.fields, record) if appropriate_verbosity?(:record)
+      adapter.reset_context
```
mbacovsky: Why it is necessary to remove the :fields from context?

mbacovsky:
@ofedoren, well done!. I like how it works and that the logic is kept quite separated in the adapter. I've left a few comments inline. A few more here:
- maybe it is time to uncover hammer-cli-foreman part, feel free to mark it [WIP] if not ready
- regarding THIN and request_params - did you tried to handle it in an option source? It could be possible to avoid changes in `request_params`
- regarding THIN and marking params explicitly - did you consider using SEARCHABLES to distinguish the right options? I'm not sure how the thin fields are defined in API and if they match to what we have in the searchables. Also we should have a plan for hammer-cli-katello
- currently it is not visible for a user what fields are in which sets. I'd love to see some help extension when --fields is defined on the command showing e.g. table like:
```
Options:
  --fileds    FIELDS  Show specified fileds or predefined filed sets only. (See below)
 
...

Predefined field sets:
+--------+-----+---------+------+
| Fields | ALL | DEFAULT | THIN |
+--------+-----+---------+------+
| Id     |  x  |    x    |  x   |
| Name   |  x  |    x    |  x   |
| Param1 |  x  |    x    |      |
| Param2 |  x  |         |      |
+--------+-----+---------+------+
```

lib/hammer_cli/output/output.rb
ofedoren: I do it to be able to reset fields selection in case of using hammer in shell mode.

lib/hammer_cli/options/option_definition.rb
ofedoren: If I leave the cleanup then the options hash for the `option` in predefined options will be empty, which makes the predefined option unusable for other commands.

ofedoren:
@mbacovsky, updated. See inline comments, also:

[Updated docs](https://github.com/theforeman/hammer-cli/blob/ce55a3afc949477263a5b25f8f8369f744dc60a2/doc/creating_commands.md#predefined-options)

As you suggested, I use now option sources instead of changing `request_params` directly, see https://github.com/theforeman/hammer-cli-foreman/pull/407

But still there is a problem: I'm afraid we need to specify `'THIN'` set explicitly for `Id` and `Name` fields in all the command output definitions (in plugins as well).

mbacovsky:
@ofedoren I'm okay with adding the help extension in separate PR. Otherwise it looks good now.
For the THIN issue - I'd really like to avoid setting it manually. I have an idea:
in the command we have `self.output` method for output definition. We could extend it in foreman commands with something like:
```ruby
output_definition.update_field_set('THIN', searchables.for(resource) + ['id])
```
`update_field_set` would iterate over the fields in definition and update the set on the matching fields.
Would it make sense?

ofedoren:
@mbacovsky, updated. i've added the `update_field_set` method as you suggested ([docs](https://github.com/theforeman/hammer-cli/blob/3ccd557a588da40de9839f828e3c12848ec67681/doc/commands_modification.md)).

The usage of it see: https://github.com/theforeman/hammer-cli-foreman/pull/407

lib/hammer_cli/output/field_filter.rb
```diff
@@ -1,21 +1,87 @@
...
+          field.display?(field_data)
+        end
+      end
+    end
+
+    def include_by_label?(labels, label)
+      return true if labels.include?(label)
+
+      labels.each do |l|
+        return true if l.start_with?("#{label}/") || label.start_with?(l)
```
mbacovsky: @ofedoren finally I was able to figure out how this is supposed to work and fixed the condition to autocomplete with asterisk `l.start_with?("#{label}/") || label.match(%r{^#{l.gsub(/\*/, '.*')}(|\/.*)$})`. I simplified it a bit also and have
```ruby
   def include_by_label?(labels, label)
      labels.any? do |l|
        l.start_with?("#{label}/") || label.match(%r{^#{l.gsub(/\*/, '.*')}(|\/.*)$})
      end
    end
```
Now it works with:
- `host info --id 4 --fields 'id,name,network'`
- `host info --id 4 --fields 'id,name,network*'`
- `host info --id 4 --fields 'id,name,network/domain'`

What do you think?

tstrachota:
Congratulations @ofedoren :beers:! It was very close to celebrate a one-year anniversary of this PR! :birthday: :hammer: :slightly_smiling_face:

PR #277, opened 2018-06-18
Fixes #21633 - Allow modification of additional help texts

ofedoren:
Currently the help string builds on the fly to actual string, so it is not so easy to change it externally. I suggest to build a sort of skeleton or some structure with definitions as it is with command's `output_definition` at first and build the help string at the end of all of the extensions/modifications.

Even though help now builds via `help.definition.build_string`, I left everything in `help/text_builder` for backwards compatibility, which leads to some code duplications in `help/definition`.

Here the [docs](https://github.com/ofedoren/hammer-cli/blob/feat-21633-modif-help-text/doc/help_modification.md).

### DIFF
--- a/lib/hammer_cli/help/definition.rb
+++ b/lib/hammer_cli/help/definition.rb
@@ -1,0 +1,60 @@
+require_relative 'definition/abstract_item'
+require_relative 'definition/text'
+require_relative 'definition/list'
+require_relative 'definition/section'
+
+module HammerCLI
+  module Help
+    class Definition < Array
+      def build_string
+        @out = StringIO.new
+        each do |item|
+          next unless item.is_a? AbstractItem
+          @out.puts unless first_print?
+          @out.puts item.build_string
+        end
+        @out.string
+      end
+
+      def find_item(item_id)
+        self[item_index(item_id)]
+      end
+
+      def at(path)
+        return super(path) if path.is_a? Integer
+        return self if path.empty?
+        expand_path(path)
+      end
+
+      def insert_definition(mode, item_id, definition)
+        index = item_index(item_id)
+        index += 1 if mode == :after
+        delete_at(index) if mode == :replace
+        definition.each_with_index do |item, offset|
+          insert(index + offset, item)
+        end
+      end
+
+      private
+
+      def expand_path(path)
+        path = [path] unless path.is_a? Array
+        item = find_item(path[0])
+        return item if path[1..-1].empty?
+        item.definition.at(path[1..-1])
+      end
+
+      def first_print?
+        @out.size.zero?
+      end
+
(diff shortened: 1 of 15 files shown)


mbacovsky:
For the first rough check the code looks well built and I appreciate the help and tests. Before I get to detailed  review I'd like to rise my concerns with the design.
@ofedoren, @tstrachota I'm a bit concerned about the flexibility we provide and that with plugins it could easily get out of hands. Namely I'm concerned about possibility to remove and change content and addressing the content by index. This could easily lead to removal of wrong content by plugins.

I'd suggest to allow only addition to the end of the definition items that has an `id` (section and list). That should provide enough control to change update the help. As an user I'd prefer information "With plugin xy it is not possible to do abc" over just removing "abc" from the help. This provides more information about why certain feature is not available and what caused that.

Would that be reasonable change to request?

ofedoren:
@tstrachota, @mbacovsky updated.

I refactored a bit previous version as you suggested with some additional changes. Also I removed the possibility to address the content by index. `delete` and `replace` methods are replaced with `insert`. In general modification of additional help text now works similar to [modification of output fields](https://github.com/theforeman/hammer-cli/pull/275).

[Updated docs](https://github.com/theforeman/hammer-cli/blob/60beac5e97c937eaca9a5844fb71b33f6effc26d/doc/help_modification.md).

lib/hammer_cli/help/definition.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      def first_print?
+        @out.size.zero?
+      end
+
+      def item_index(item_id)
+        index = find_index do |item|
+          item.id == item_id
+        end
+        raise ArgumentError, "Help item #{item_id} not found" if index.nil?
```
mbacovsky: It may be better to wrap the item_id in apostrophes. When `item_id` is the section label it would be easier to find in the message.

It may be better user experience to catch the error at some point, print a warning that the attempt to extend the help failed on missing item_id and log the backtrace but go on and print as much help as we can. That way outdated plugin trying to extend the help couldn't break the execution.

mbacovsky:
[test hammer-cli]

mbacovsky:
[test hammer]

mbacovsky:
@ofedoren that looks pretty well now. Is it a final version that can be merged?

mbacovsky:
The failing tests are not related, I'll rerun later.

mbacovsky:
@tstrachota could you revisit, please?

ofedoren:
@mbacovsky yes, if there are no additional proposes, it is the final version.

mbacovsky:
[test hammer]

mbacovsky:
Thanks @ofedoren, @tstrachota!

PR #280, opened 2018-07-31
Fixes #24488 - restore help building

ares:
found while I was adding a new command, the resource seems to be class method only

### DIFF
--- a/lib/hammer_cli/apipie/resource.rb
+++ b/lib/hammer_cli/apipie/resource.rb
@@ -7,7 +7,7 @@ module HammerCLI::Apipie
     end
 
     def resource
-      self.class.resource || self.parent_command.class.resource
+      self.class.resource || parent_command_resource
     end
 
     def action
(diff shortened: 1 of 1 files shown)


lib/hammer_cli/apipie/command.rb
```diff
@@ -38,7 +38,7 @@ def execute
 
     def help
       help_str = super
-      if !resource || (!action.nil? && !resource.has_action?(action))
+      if !self.class.resource || (!action.nil? && !self.class.resource.has_action?(action))
```
mbacovsky: I think the original version is correct. Instance method `resource` is defined here https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/apipie/resource.rb#L9 and is properly called. The problem seems to be in https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/apipie/resource.rb#L10 where the `self.parent_command.class.resource` is not defined (MainCommand is not successor of ApipieCommand and has no resource). The proper fix should likely be to check for existence of the resource method there.

I've update the issue with explanation of what is causing the error.

mbacovsky:
:+1: This fix LGTM now. Thanks @ares!

PR #281, opened 2018-09-06
Fixes #17010 - Hammer JSON output uses inconsistent capitalization

ofedoren:
Since there are people which prefer different capitalization types, I suggest to let the user to specify which one they want in `cli_config.yml`.

This fix also fixes capitalization of YAML output.

### DIFF
--- a/config/cli_config.template.yml
+++ b/config/cli_config.template.yml
@@ -10,6 +10,9 @@
   #:mark_translated: false
   # Hide headers from output of list actions
   # :no_headers: true
+  # Choose capitalization for JSON/YAML output.
+  # One of: 'downcase', 'capitalize', 'upcase'
+  # :capitalization: 'downcase'
 
 # Enable/disable color output of logger in Clamp commands
 :watch_plain: false
--- a/lib/hammer_cli/context.rb
+++ b/lib/hammer_cli/context.rb
@@ -1,16 +1,13 @@
 require 'hammer_cli/defaults'
 
 module HammerCLI
-
   def self.context
     @context ||= {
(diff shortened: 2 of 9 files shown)


ofedoren:
@tstrachota updated.

Also it's possible now to switch from e.g. `downcase` to current default with empty string `''` or blank value (`nil` will be converted to string and will cause warning). Commenting out works as well.

tstrachota:
I like using `''` to turn the functionality off. I did some testing and it works well in most of the cases. For some reason some reference fields don't get capitalized. Try for example `subnet info` with upcase:

```
hammer --output=json subnet info --id 1
{
  "ID": 1,
  "NAME": "tstracho-laptop",
  "PROTOCOL": "IPv4",
  "SMART PROXIES": {
    "DNS": null,
    "TFTP": {
      "name": "tstracho-laptop",
      "id": 2,
      "url": "http://192.168.122.1:8000"
    },
    "DHCP": null
  },
 #...
```
do you have a clue why the keys inside tftp remain in lowcase?

ofedoren:
@tstrachota I found a [fix](https://github.com/theforeman/hammer-cli/pull/281/files#diff-4e7adcf99edc71f00145de9875097aa9R54) for what you described, but I'm not sure if this will fix another nested layer (probably not).

tstrachota:
Thank you @ofedoren your fix for references is good enough for now. This PR is major improvement that covers most of the cases and I don't want to block it any more. :+1:

tstrachota:
BTW this is nice candidate for the next community demo.

PR #284, opened 2018-09-13
Fixes #19191 - Output formatter applies for selected adapter only

ofedoren:
Now it is possible to apply an output formatter on selected adapters only.

To use this feature, define `applicable_adapters` **OR** `not_applicable_adapters` method in your formatter like this:

```ruby
class MyFormatter < HammerCLI::Output::Formatters::FieldFormatter
  def applicable_adapters
    [HammerCLI::Output::Adapter::Table]
  end

  def tags
    # tags
  end
  
  def format(data, field_params = {})
    # logic
  end
end
```

Similar to apply on all the adapters except a particular one:

```ruby
class MyFormatter < HammerCLI::Output::Formatters::FieldFormatter
  def not_applicable_adapters
    [HammerCLI::Output::Adapter::Table]
  end

  def tags
    # tags
  end
  
  def format(data, field_params = {})
    # logic
  end
end
```

**NOTE:** If none of the methods defined, everything has default behaviour.

### DIFF
--- a/lib/hammer_cli/output/adapter/abstract.rb
+++ b/lib/hammer_cli/output/adapter/abstract.rb
@@ -3,7 +3,7 @@ module HammerCLI::Output::Adapter
   class Abstract
 
     def tags
-      []
+      [:abstract]
     end
 
     def initialize(context={}, formatters={})
--- a/lib/hammer_cli/output/adapter/base.rb
+++ b/lib/hammer_cli/output/adapter/base.rb
@@ -5,7 +5,7 @@ module HammerCLI::Output::Adapter
     LABEL_DIVIDER = ": "
 
     def tags
-      [:flat, :screen]
+      [:base, :flat, :screen]
     end
 
     def print_record(fields, record)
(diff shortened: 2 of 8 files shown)


mbacovsky:
@ofedoren looks good! Could you cover this in tests?

Did you consider using adapter names (under which they are registered) - `:table`, `:csv` instead of the class name? This way the formatter filtering would still work if e.g. the table adapter implementation is replaced. Just an idea, not sure if that is a good thing yet.

mbacovsky:
@ofedoren I've also re-checked the original issue and think this approach duplicates the tags. I'll comment in the issue.

ofedoren:
@mbacovsky, updated.

Now it uses tags for comparison.

ofedoren:
blocked by https://github.com/theforeman/hammer-cli/pull/300

mbacovsky:
@ofedoren, the more I think about this the more I tend to keep just #300  and close this PR and the #19191 issue as won't fix or duplicate. The only use-case mentioned was fixed by 300 and I think it is better practice to add new "tags" then make formatters bound to specific adapters. Makes sense?

ofedoren:
@mbacovsky, I didn't even notice, that #300 fixed that use-case :) As shows that PR, we can add another formatter or "tags", so I'll close this PR with the issue as duplicate.

PR #285, opened 2018-09-18
Fixes #21359 - Hammer no longer supports ruby < 2.0

ofedoren:
Cherry-picked from https://github.com/theforeman/hammer-cli/pull/269 and added some changes.


ofedoren:
@tstrachota, updated. I returned `String#format` method, since it also does interpolation when a hash with parameters has keys as strings (ruby's `String#%` with hashes works only for keys as symbols).

mbacovsky:
@ofedoren thanks for keeping the repo tidy! The patch looks good to go.
@tstrachota would you mind revisiting this?

tstrachota:
Merged, thanks @ofedoren !

PR #293, opened 2018-10-26
Fixes #25307 - Add ability to disable defaults

mbacovsky:
Presence of Defaults can be set in cli_config.yml with
:use_defaults: true/false or on CLI with --use-defaults
--no-use-defaults respectively. both ways can be combined.
The value is propagated in the context as :use_defaults.
The defaults are enabled by default.

Hammer settings class was chenged so that it can be the single
source of default values which has three benefits
 - we don't have to fallback to default with every use of the value
 - we can keep the values commented out in config which makes it easier
to changelater
 - there is single place to look for the default setting

### DIFF
--- a/bin/hammer
+++ b/bin/hammer
@@ -84,7 +84,6 @@ HammerCLI::Settings.load({
     :interactive => preparser.interactive,
     :debug => preparser.debug?,
     :no_headers => preparser.no_headers?,
-    :use_defaults => preparser.use_defaults?,
     :reload_cache => preparser.reload_cache?,
     :verify_ssl => preparser.verify_ssl,
     :ssl_ca_file => preparser.ssl_ca_file,
--- a/bin/hammer
+++ b/bin/hammer
@@ -13,6 +13,7 @@ require 'hammer_cli/options/normalizers'
 # Option descriptions are never displayed and thus do not require translation.
 class PreParser < Clamp::Command
   option ['-v', '--[no-]verbose'], :flag, _('Be verbose (or not). True by default')
+  option ['--[no-]use-defaults'], :flag, _('Enable/disable stored defaults. Enabled by default')
   option ['-q', '--quiet'], :flag, _('Completely silent')
   option ["-d", "--debug"], :flag, "show debugging output"
   option ["-c", "--config"], "CFG_FILE", "path to custom config file" do |path|
@@ -83,6 +84,7 @@ HammerCLI::Settings.load({
     :interactive => preparser.interactive,
     :debug => preparser.debug?,
     :no_headers => preparser.no_headers?,
+    :use_defaults => preparser.use_defaults?,
     :reload_cache => preparser.reload_cache?,
     :verify_ssl => preparser.verify_ssl,
     :ssl_ca_file => preparser.ssl_ca_file,
@@ -91,6 +93,8 @@ HammerCLI::Settings.load({
     :ssl_client_key => preparser.ssl_client_key,
     :ssl_with_basic_auth => preparser.ssl_with_basic_auth?
   }})
+
+HammerCLI::Settings.load({:use_defaults => preparser.use_defaults?}) unless preparser.use_defaults?.nil?
 
 if HammerCLI::Settings.get(:ui, :mark_translated)
   include HammerCLI::I18n::Debug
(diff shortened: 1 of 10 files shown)


lib/hammer_cli/settings.rb
```diff
@@ -59,9 +64,16 @@ def self.path_history
       @path_history
     end
 
+    def self.default_values
```
tstrachota: How about renaming it to `default_settings` ?
It's probably just because of this PR's context, but the name associated me different functionality.

lib/hammer_cli/defaults.rb
```diff
@@ -100,9 +100,13 @@ def switch_to_name(opt)
...
-    @defaults ||= Defaults.new(HammerCLI::Settings.settings[:defaults])
-
+    return @defaults if @defaults
+    settings = HammerCLI::Settings
+    default_values = settings.get(:use_defaults) ? settings.get(:defaults) : {}
+    @defaults = Defaults.new(default_values)
   end
 
-  HammerCLI::MainCommand.subcommand "defaults", _("Defaults management"), HammerCLI::DefaultsCommand
+  if HammerCLI::Settings.get(:use_defaults)
```
tstrachota: I think we should keep the command in. I find it confusing that a command would disappear when you pass `--no-defaults`

lib/hammer_cli/defaults.rb
```diff
@@ -100,9 +100,13 @@ def switch_to_name(opt)
   end
 
   def self.defaults
-    @defaults ||= Defaults.new(HammerCLI::Settings.settings[:defaults])
-
+    return @defaults if @defaults
+    settings = HammerCLI::Settings
+    default_values = settings.get(:use_defaults) ? settings.get(:defaults) : {}
+    @defaults = Defaults.new(default_values)
```
tstrachota: Maybe it's just me, but this seems difficult to read.
How about:
```ruby
@defaults ||= Defaults.new(default_values_from_settings)

private
def default_values_from_settings
  HammerCLI::Settings.get(:use_defaults) ? HammerCLI::Settings.get(:defaults) : {}
end
```

lib/hammer_cli/defaults.rb
mbacovsky: @tstrachota Removing commands was easy way to prevent the `defaults.yaml` from rewriting when user stores new value with defaults disabled. I can make it fail on attempt to store the defaults. Would that be better?

lib/hammer_cli/defaults.rb
mbacovsky: @tstrachota it's a way better! You were awarded by the `rubocop of the day` trophy. :robot:

lib/hammer_cli/defaults.rb
ofedoren: I'd vote for fail on attempt and showing user a some kind of warning?

bin/hammer
```diff
@@ -92,6 +94,8 @@ HammerCLI::Settings.load({
     :ssl_with_basic_auth => preparser.ssl_with_basic_auth?
   }})
 
+HammerCLI::Settings.load({:use_defaults => preparser.use_defaults?}) unless preparser.use_defaults?.nil?
```
ofedoren: Isn't it a duplication of https://github.com/theforeman/hammer-cli/pull/293/files#diff-aca6a835822ab9b99782be0a5f01f8ddR87 ? If it's a shorcut for `Settings.get(:_params, :use_defaults)`, why keep it under `_params` then?

mbacovsky: @ofedoren Exactly. I had a reason to have it there but it was bad idea indeed. Removed.

mbacovsky:
@tstrachota, @ofedoren:  comments addressed in separate commit. Is it better?
Please squash before merging ;)

lib/hammer_cli/defaults.rb
tstrachota: IMHO it's totally fine to save and list defaults even `--no-use-defaults` is passed. The option should just deactivate using the defaults in requests.

tstrachota:
Looks ok now. Just one question regarding the design @mbacovsky:
Have you considered just adding equivalent of `unless context[:use_defaults]` here https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/abstract.rb#L250 ? It seems as the most straightforward solution to me. It would probably also allow keeping the defaults commands' functionality on even with disabled defaults.

mbacovsky:
@tstrachota updated. I removed the restriction to save the defaults when disabled. I also moved the logic to the option source.

mbacovsky:
@tstrachota updated. Last version?

tstrachota:
Looks good and works as expected now. Thanks @mbacovsky

PR #295, opened 2018-11-14
Fixes #25438 - hammer -h does work with --output json, csv, yaml

ofedoren:


ofedoren:
@tstrachota, some tests are added.

Also, I think, I've found and fixed the cause of that error: https://github.com/theforeman/hammer-cli/pull/295/files#diff-ea442e923f418470f2c3930bae635023L53

tstrachota:
Great. Yeah, I found that place too, thanks. I feel more confident when we know where the nil was coming from :)

tstrachota:
Merged, thanks @ofedoren

PR #296, opened 2018-12-05
Update rpm names to use tfm

akofink:


akofink:
@tstrachota care to review?

tstrachota:
Thanks @akofink !

PR #297, opened 2018-12-18
Refs #25724 - Fixed instructions in i18n docs

tstrachota:
The previous instructions didn't work with bundler when make was executed from ./locale/


mbacovsky:
@tstrachota looks good, thanks for the update! :memo:
:+1:

PR #298, opened 2019-03-07
Fixes #26120 - Extended capabilities for list type options

ofedoren:
Changes:
  - Added JSON support for `List` normalizer.
  - The option mentioned in the issue has [NestedValidator](https://github.com/Apipie/apipie-rails#nestedvalidator), so expected input is an array with hash objects. The plain `List` normalizer is good for simple lists (arrays). So, I've decided to create a mixed normalizer (`List` + `KeyValueList`) to handle this kind of options.
  - The help now contains `Option details` secion for each option of that sort, where user can see expected input generated from apidoc.
  - Added notification of JSON support to the help for each option that supports this kind of input.

Known bugs:
  - Because of simple regex in `KeyValueList` normalizer, there is no native support for more complicated input, like array of hashes or hash with an array as a value. That's why JSON support was added.

### DIFF
--- a/test/unit/options/normalizers_test.rb
+++ b/test/unit/options/normalizers_test.rb
@@ -14,21 +14,21 @@ describe HammerCLI::Options::Normalizers do
   end
 
   describe 'default' do
-  
+
     let(:formatter) { HammerCLI::Options::Normalizers::Default.new }
-  
+
     it "should not change any value" do
       formatter.format('value').must_equal 'value'
     end
-  
+
     it "should not change nil value" do
       formatter.format(nil).must_be_nil
       end
-    
+
     it "has empty description" do
       formatter.description.must_equal ''
     end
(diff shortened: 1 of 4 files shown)


test/unit/options/normalizers_test.rb
```diff
@@ -69,8 +69,85 @@
     it "should catch quoting errors" do
       proc { formatter.format('1,"3,4""s') }.must_raise ArgumentError
     end
+
+    it "should accept and parse JSON" do
+      formatter.format("{\"name\":\"bla\", \"value\":1}").must_equal(
```
mbacovsky: @ofedoren, wouldn't the following be more readable?
```ruby
      value = {'name' => 'bla', 'value' => 1}
      formatter.format(value.to_json).must_equal(value)
```

PR #301, opened 2019-04-02
Fixes #24964 - Document how to deprecate commands

ofedoren:
Except documentation the fix adds a way to mark a command as deprecated in its definition (in case of subcommand autoloading):

```ruby
class Command < HammerCLI::AbstractCommand
  deprecation 'This command is deprecated.'
  # ...
  def execute
    # ...
  end
end
```

### DIFF
--- a/doc/creating_commands.md
+++ b/doc/creating_commands.md
@@ -435,7 +435,34 @@ HammerCLI::MainCommand.lazy_subcommand(
 )
 ```
 
+### Deprecated commands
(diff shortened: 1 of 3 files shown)


mbacovsky:
@ofedoren this looks good! I appreciate the added docs :beer:.
I was curious why the option and command does not use consistent names e.g. `deprecation_warning`?

ofedoren:
@mbacovsky. good question :) I've changed names, so there only `warning` is used.

PR #302, opened 2019-04-29
Fixes #26703 - Make sure list opts return a list

mbacovsky:


mbacovsky:
@ofedoren could you please review. This is regression affecting many *_ids options.

mmoll:
a 0.17.1 release with this would be nice

PR #308, opened 2019-06-14
Fixes #26960 - The --fields option has set help

ofedoren:


lib/hammer_cli/output/definition.rb
```diff
@@ -54,8 +54,84 @@ def empty?
       @fields.empty?
     end
 
+    def field_sets
+      nested_fields_sets(@fields).uniq.sort
+    end
+
+    def sets_table
+      fields_col_size = average_label_length
```
mbacovsky: What was the reason for the average length here? It tends to produce unnecessarily narrow tables with part of the text hidden behind the dots.
```
  +------------+-----+---------+------+
  |   Fields   | ALL | DEFAULT | THIN |
  +------------+-----+---------+------+
  | Id         |  x  |    x    |  x   |
  | Name       |  x  |    x    |  x   |
  | Operati... |  x  |    x    |      |
  | Host Group |  x  |    x    |      |
  | IP         |  x  |    x    |      |
  | MAC        |  x  |    x    |      |
  | Global ... |  x  |    x    |      |
  | Organiz... |  x  |         |      |
  | Location   |  x  |         |      |
  | Additio... |  x  |         |      |
  | Content... |  x  |    x    |      |
  | Lifecyc... |  x  |    x    |      |
  +------------+-----+---------+------+
```
Wouldn't it be better to take terminal width and fit the table in and reduce the content only when it doesn't fit?

mbacovsky: As per our discussion we decided to use max_label_length instead.

ofedoren:
@mbacovsky, thanks, updated :)

mbacovsky:
We decided to address the table refactoring in separate PR as it may involve steps out of scope of this PR. It is tracked in https://projects.theforeman.org/issues/27318

PR #309, opened 2019-07-08
Fixes #27237 - List items in help with customization

ofedoren:


doc/help_modification.md
```diff
@@ -1,6 +1,35 @@
...
+  h.text(content, options = {})
+  # Add a notification to the help output (LABEL: content).
+  # content - String - text to be shown
+  # options - Hash   - options (same as #text, label:)
+  h.note(content, options = {})
+  # Add a list of items to the help output.
+  # items   - Array - array of items to be shown ([ ['Key', 'Value', { option: value }] ])
+  # options - Hash  - same as #text
+  # block   - optional block with custom modifications of each item in the list (must return modified item)
+  h.list(items, options = {}) { |item| ... }
```
mbacovsky: I'm missing description of the {option: value} at the item level. From the code I'm seeing it can be `bold:`. Anything else? Also this is a bit confusing as it produces Key/Value list by default. Any examples how to format plain list of values properly indented?

Also what is the purpose of modification of the data in the block? Wouldn't it be more readable if we prepare the data before we try to format them? I'd see the benefit of the block if the items are expected to be list of strings and the block would do serialization of the items otherwise. e.g. 

```ruby
h.list(['a', 'b', 'c'])
```
or
```ruby
h.list([['key', 'value', true], ['foo', 'bar', false]]) { |item| "#{item[2] ? bold(item[0]) : item[0] }: #{item[1]}" }
```
Am I missing something?

doc/help_modification.md
```diff
@@ -1,6 +1,35 @@
...
+# Modify an existing help
+
+### Help items
+Before modifying help, it's good to know which help items are available to use.
+There is a list with currently available items:
+```ruby
+Command.extend_help do |h|
+  # Add a simple text to the help output.
+  # content - String - text to be shown
+  # options - Hash   - options (id:, richtext:)
```
mbacovsky: `richtext:` here means 'bold'?

mbacovsky: by richtext I'd probably expect something like https://github.com/piotrmurach/tty-markdown ;) Didn't you consider it? Might be next PR candidate ;)

ofedoren:
@mbacovsky, updated as per our discussion :)

doc/help_modification.md
```diff
@@ -1,6 +1,36 @@
...
+  h.text(content, options = {})
+  # Add a notification to the help output (LABEL: content).
+  # content - String - text to be shown
+  # options - Hash   - options (same as #text, label:)
+  h.note(content, options = {})
+  # Add a list of items to the help output.
+  # items   - Array - array of items to be shown. Each item can be an array.
+  #   Example: h.list(['a', 'b', 'c']) => "a    b    c\n"
```
mbacovsky: shouldn't that be `"a\n    b\n    c\n"` ?

doc/help_modification.md
```diff
@@ -1,6 +1,36 @@
...
+  # Add a simple text to the help output.
+  # content - String - text to be shown
+  # options - Hash   - options (id:, richtext:)
+  h.text(content, options = {})
+  # Add a notification to the help output (LABEL: content).
+  # content - String - text to be shown
+  # options - Hash   - options (same as #text, label:)
+  h.note(content, options = {})
+  # Add a list of items to the help output.
+  # items   - Array - array of items to be shown. Each item can be an array.
```
mbacovsky: Can we document also the item options? How about "Each item can be either string or an array of `['Key', 'Value', bold: true/false]`. Item options are optional."

ofedoren:
@mbacovsky, thanks! Updated.

PR #312, opened 2019-08-28
Fixes #27728 - Add new bash completion

ofedoren:


ofedoren:
@mbacovsky, updated. I've noticed that we don't load settings from hammer-cli/config/cli_config.template.yml file. See my addition: https://github.com/theforeman/hammer-cli/blob/f57a352ad674da263f3c2434878843abbd7fc8bd/lib/hammer_cli/settings.rb#L20

@shiramax, If you run ./bin/hammer-complete you will see that the hammer will be initialized fully only if `~/.cache/hammer_completion.yml` file is missing. Also, on dev env if you want to speed-up autocompletion, add $path_to_hammer_bin on the first place in $PATH environment variable.

ofedoren:
@shiramax, @mbacovsky, I think it's ready to be merged.

Also, the completion time is cca 0.17s.

bin/hammer-complete
```diff
@@ -0,0 +1,55 @@
...
+   .each do |file|
+    File.open(file) do |f|
+      config = YAML.load(f)
+      cache_file = config[:completion_cache_file] || cache_file
+    end
+  end
+  cache_file
+end
+
+completion_cache_file = cache_file || '~/.cache/hammer_completion.yml'
```
mbacovsky: @ofedoren it may spare some cycles if we make it more opinionated and define it as `ENV['HAMMER_COMPLETION_CACHE'] || '~/.cache/hammer_completion.yml'`. No idea how much time we spend with the config loading and parsing. We could do similarly for the hammer command name (`HAMMER`).

bin/hammer-complete
```diff
@@ -0,0 +1,55 @@
...
+  require 'hammer_cli'
+  `#{HAMMER} prebuild-bash-completion`
+end
+
+require 'hammer_cli/bash/completion'
+
+dict = HammerCLI::Bash::Completion.load_description(completion_cache_file)
+
+# regenerate and reload cache if it expired
+sha1sum = `sha1sum #{dict['expire']['file']}` if File.exist?(dict['expire']['file'])
```
mbacovsky: When we expire the cache based on apipie cache couldn't we remove the completion cache with apipie cache update? That way we could avoid re-computing sha of the long file with every <tab>. Would that be possible @ofedoren?

ofedoren:
@mbacovsky, updated. I've removed sha computing, so the new completion cache is created when reloading apipie cache or if the completion cache file doesn't exist. Also removed using of `which` command.

bin/hammer-complete
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env ruby
+
+require 'English'
+@project_root = File.expand_path('../..', __FILE__)
+$LOAD_PATH.unshift(File.join(@project_root, 'lib'))
+HAMMER = ENV['HAMMER'] || File.join(@project_root, 'bin/hammer')
+
+require 'yaml'
+
+completion_cache_file = ENV['HAMMER_COMPLETION_CACHE'] || '~/.cache/hammer_completion.yml'
```
xprazak2: `cli_config.yml` is loaded into ENV and we pull the value for cache location from there or use default, right?

xprazak2:
Works nicely for me, just one thing - when I was randomly testing commands, I got a strange value completed for the os command, just for the os-params-attrs:

os-params-attrs

xprazak2:
Also test failures seem related.

ofedoren:
@xprazak2, updated. The strange value is fixed according to our discussion and the tests are fixed :)

xprazak2:
Thanks everyone!

tbrisker:
Build is now failing: https://community.theforeman.org/t/hammer-cli-master-package-release-29-failed/16777

xprazak2:
Packaging fix: https://github.com/theforeman/foreman-packaging/pull/4539

PR #313, opened 2019-09-05
Fixes #27595 - Better prompts for missing arguments

ofedoren:
Changes prompts like
```
Missing arguments for 'foreman_virt_who_configure_config[hypervisor_username]'
```
to (as per help)
```
Missing arguments for '--hypervisor-username'
```

### DIFF
--- a/lib/hammer_cli/exception_handler.rb
+++ b/lib/hammer_cli/exception_handler.rb
@@ -138,7 +138,16 @@ module HammerCLI
     end
 
     def handle_apipie_missing_arguments_error(e)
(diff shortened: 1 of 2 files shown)


xprazak2:
Works, could you add a simple test(s) for this please? Regex can cause unpleasant surprises...

ofedoren:
@xprazak2, thanks for reviewing! :)

I've added some tests and also adjusted the fix to support (simply) the more nested arguments.

xprazak2:
Nice, I think this is ready.

PR #314, opened 2019-09-05
Fixes #27318 - Extract table generator into reusable component

ofedoren:
This PR introduces a new component -- `HammerCLI::Output::Generators::Table`, which generates a table based on provided `columns` and `data`.

Basic usage:
```ruby
# can be an array of strings or a hash e.g. { column_label: params }
columns = ['First col', 'Second col'] || { 'First col' => { width: 20 }, 'Second col' => { width: 20 } }
# prepared data based on your case. Should be an array of hashes e.g. { col_label: col_data }
data = [
  { 'First col' => 1, 'Second col' => 2 },
  { 'First col' => 3, 'Second col' => 4 }
]
# hash with options
options = { no_headers: false }

table_gen = HammerCLI::Output::Generators::Table.new(columns, data, options)
# to get the whole table
table_gen.result # =>
#----------------|---------------------
#FIRST COL       | SECOND COL
#----------------|---------------------
# 1              | 2                      
# 3              | 4      

# or you can get specific parts of the table
table_gen.header
table_gen.body
table_gen.footer
```

### DIFF
--- a/lib/hammer_cli/output.rb
+++ b/lib/hammer_cli/output.rb
@@ -1,9 +1,9 @@
 require File.join(File.dirname(__FILE__), 'output/output')
 require File.join(File.dirname(__FILE__), 'output/fields')
 require File.join(File.dirname(__FILE__), 'output/formatters')
+require File.join(File.dirname(__FILE__), 'output/generators')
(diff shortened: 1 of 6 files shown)


xprazak2:
Nice refactoring, works as expected.

PR #315, opened 2019-09-09
Fixes #21590 - Added error to wrong --output

yifatmakias:


lib/hammer_cli/output/output.rb
```diff
@@ -36,6 +36,14 @@ def print_collection(definition, collection)
       adapter.reset_context
     end
 
+    def print_output_type
+      if self.class.adapters.has_key? context[:adapter].to_sym
+        $stderr.puts _("Using '#{context[:adapter]}' adapter")
```
ofedoren: I'd prefer method `warn` that we're already [using](https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/utils.rb#L81) instead of using directly `$stderr.puts`. What do you think?

lib/hammer_cli/main.rb
```diff
@@ -47,8 +47,8 @@ class MainCommand < AbstractCommand
       :context_target => :interactive
     option ["--no-headers"], :flag, _("Hide headers from output")
     option ["--csv"], :flag, _("Output as CSV (same as --output=csv)")
-    option ["--output"], "ADAPTER", _("Set output format. One of [%s]") %
-      HammerCLI::Output::Output.adapters.keys.join(', '),
+    option ["--output"], "ADAPTER", _("Set output format."),
```
ofedoren: The option description must not end with a dot.

lib/hammer_cli/main.rb
```diff
@@ -47,8 +47,8 @@ class MainCommand < AbstractCommand
...
+           format: HammerCLI::Options::Normalizers::Enum.new(HammerCLI::Output::Output::adapters.keys.map(&:to_s)),
```
ofedoren: Prefer `Class.method` instead of `Class::method`.

lib/hammer_cli/main.rb
yifatmakias: Thanks, I changed it.

lib/hammer_cli/main.rb
yifatmakias: Thanks, I changed it.

ofedoren:
[test]

ofedoren:
[test hammer]

PR #316, opened 2019-09-20
Fixes #27899 - Allow schema building for custom options

ofedoren:


shiramax:
@ofedoren what do you think about this help:

Option details:
  Following parameters accept format defined by its schema (bold are required, parameter_name and value must be a string):

  --typed-parameters  "name=<parameter_name>\,value=<parameter_value>\,parameter_type=[string || int ...]]  "

ofedoren:
@shiramax, thanks for your effort to make UX better!

I agree that it will help users, but what if we go further and show something even more verbose?
```
  Following parameters accept format defined by its schema (bold are required; <> contain acceptable type; [] contain acceptable value):

  --typed-parameters  "name=<string>\,value=<string>\,parameter_type=[string|boolean|integer|real|array|hash|yaml|json]\,hidden_value=[true|false|1|0], ... "
```

Please tell me what do you think about that :)

shiramax:
@ofedoren , looks good !

ofedoren:
@shiramax, updated :) I've added a way to show mentioned help, so when you have time please check it (with https://github.com/theforeman/hammer-cli-foreman/pull/448 for example) and tell me if it's better.

PR #320, opened 2019-12-17
Fixes #28503 - Allow column max width more than 80

ofedoren:


apuntamb:
Can we merge this as [hammer-cli-foreman#468](https://github.com/theforeman/hammer-cli-foreman/pull/468) is dependent on this.

shiramax:
thanks @ofedoren, @chris1984 and @apuntamb !

PR #322, opened 2020-01-17
Fixes #28793 - Better promts for missing arguments

ofedoren:


xprazak2:
Looks ok to me, do we have any existing command that would prompt us for missing `--opt_a` instead of `--opt-a` to tests this?

ofedoren:
@xprazak2, yes. I've tried `hammer compute-resource resource-pools --id 1`. It asks for `--cluster_id`.

xprazak2:
Thanks everyone!

PR #323, opened 2020-01-27
Fixes #28440 - Help contains squeezed options

ofedoren:
This PR introduces a way to squeeze options in help output by grouping the options in 'families'.

TODO:
  - [x] - Docs
  - [x] - Tests

Example (for better examples see https://github.com/theforeman/hammer-cli-foreman/pull/489):

```ruby
class MyCommand
  option_family(description: 'Desc for the main switch') do
    parent '--item-id', '', 'Item desc 1'
    child '--item', '', 'Item desc 2'
  end
end

# > hammer -h
# ...
# Options:
#   --item[-id]            Desc for the main switch
# ...
# > hammer full-help
# ...
# Options:
#   --item                   Item desc 2
#   --item-id                Item desc 1
# ... 
```

### DIFF
--- a/doc/creating_commands.md
+++ b/doc/creating_commands.md
@@ -173,6 +173,54 @@ Here is the list of predefined options:
   * `:fields` Expects a list with fields to show in output, see [example](creating_commands.md#printing-hash-records).
 
 
+### Option family
+Option family is the way to unify options which have the same meaning or purpose,
+but contain some differences in their definitions (e.g. the name/switch of an option).
+Mainly serves as a container for options, which purpose is to show less repetitive
+output in commands' help. Option builders use it by default.
+
+To define an option family, use the following DSL:
+```ruby
+  # options is a Hash with options for family/each defined option within it
+  option_family(options = {}) do
+    # parent is the main option. Must be single, option family can have only one parent.
+    parent switches, type, description, options
+    # child  is an additional option. Could be none or more than one. Aren't shown in the help output.
+    child  switches, type, description, options
+  end
+```
+
+##### Example
+
+```ruby
+  option_family(
+    aliased_resource: 'environment',
+    description: _('Puppet environment'),
+    deprecation: _("Use %s instead") % '--puppet-environment[-id]'
+    deprecated: { '--environment' => _("Use %s instead") % '--puppet-environment[-id]',
+                  '--environment-id' => _("Use %s instead") % '--puppet-environment[-id]'}
+  ) do
+    parent '--environment-id', 'ENVIRONMENT_ID', _(''),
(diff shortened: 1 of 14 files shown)


lib/hammer_cli/options/option_family.rb
```diff
@@ -0,0 +1,80 @@
...
+          m.switches.map { |s| s.gsub(main_switch, '') }
+        end.flatten.reject(&:empty?).sort { |x, y| x.size <=> y.size }
+        "#{main_switch}[#{suffixes.join('|')}]"
+      end
+
+      def all
+        @children + [@parent].compact
+      end
+
+      def head
```
xprazak2: I do not see this method being used.

lib/hammer_cli/options/option_family.rb
```diff
@@ -0,0 +1,80 @@
...
+          opts[:deprecated] = opts[:deprecated].select do |switch, _msg|
+            handles.include?(switch)
+          end
+        end
+        @creator.instance_eval do
+          option(switches, type, description, opts, &block)
+        end
+      end
+
+      def optionamize(name_candidate)
```
xprazak2: While I do get why the method is named like this, the name just looks strange to me. Can you think of a better name? I can share my suggestions if you'd like...

xprazak2:
Provides a base for squeezed options and seems ok, but my knowledge in this area is fairly limited.

lib/hammer_cli/options/option_family.rb
ofedoren: @xprazak2, actually we already have that method in option builders https://github.com/theforeman/hammer-cli/blob/16a30c6e93cecbbf5321f402b1abb75ca428a8b4/lib/hammer_cli/option_builder.rb#L15

But I'm open to a new better name.

xprazak2: Oh, I thought it was a new one. In that case, I think it is ok to keep.

doc/creating_commands.md
```diff
@@ -173,6 +173,54 @@ Here is the list of predefined options:
...
+Mainly serves as a container for options, which purpose is to show less repetitive
+output in commands' help. Option builders use it by default.
+
+To define an option family, use the following DSL:
+```ruby
+  # options is a Hash with options for family/each defined option within it
+  option_family(options = {}) do
+    # parent is the main option. Must be single, option family can have only one parent.
+    parent switches, type, description, options
+    # child  is an additional option. Could be none or more than one. Aren't shown in the help output.
```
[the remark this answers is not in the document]
ofedoren: switches are `'--something'` or `['--option-one', '--option-two']`. It's how it's named in Clamp.

PR #326, opened 2020-02-14
Fixes #29015 - Keep referenced resource in option options

ofedoren:


xprazak2:
It is probably better not to delete it, thanks!

PR #327, opened 2020-03-26
Fixes #29413 - Add fuzzy subcommand matching

ofedoren:
Adds fuzzy (prefix) subcommand matching as in https://github.com/tstrachota/hammer-cli-experimental.

Allows to use 'reduced' hammer commands, for example:
 - `hammer org l` instead of `hammer organization list`
 - `hammer ro s --id 1` instead of `hammer role show --id 1`
 - `hammer li l` instead of `hammer lifecycle-environment list`

If there is anthoer subcommand that starts with the same prefix, the help error is shown, e.g.:
```
> hammer host i
Error: Found more than one command.

Did you mean one of these?
	index
	info
``` 

Also adds possibility to use aliases for commands, for example:
 - `hammer host show --id 1` instead of `hammer host info --id 1`
 - `hammer host index` instead of `hammer host list`

PR that adds aliases for info/list/delete commands: https://github.com/theforeman/hammer-cli-foreman/pull/512

### DIFF
--- a/doc/creating_commands.md
+++ b/doc/creating_commands.md
@@ -454,6 +454,31 @@ Options:
     -h, --help                    print help
 ```
 
+#### Aliasing subcommands
+
+Commands can have two or more names, e.g. aliases. To support such functionality
+simple name addition could be used via `command_name` or `command_names` method:
+```ruby
+module HammerCLIHello
+
+  class SayCommand < HammerCLI::AbstractCommand
+
+    class GreetingsCommand < HammerCLI::AbstractCommand
+      command_name 'hello'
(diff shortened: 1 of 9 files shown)


ofedoren:
@ares, you might want to take a look at this :)

ares:
Woot, I won't be able to properly review anytime soon, but quick read seems ok

ofedoren:
@shiramax, done: https://github.com/theforeman/hammer-cli/blob/aa8a2d753bbeb45a3745ae7e2c3a872139b48b74/doc/creating_commands.md#aliasing-subcommands :)

PR #329, opened 2020-05-05
Fixes #29697 - hammer full-help returns correct output

ofedoren:
The issue is that even though all commands work, some of them have `Unfortunately the server does not support such operation` strings in full-help output. This is due to different initialization process of commands. This fix allows instantiated commands to see actual invocation path and to be able to see parent `resource`.


xprazak2:
Takes care of the issue for me, thanks!

PR #332, opened 2020-05-26
Update rake to its latest version

ekohl:


ekohl:
Looks like it fails and I don't have time to look into it. https://community.theforeman.org/t/insecure-version-of-rake-used-all-over-hammer/18936

PR #334, opened 2020-08-19
Fixes #30666 - add details to Installation from source doc

MariaAga:
Installation from source doc should include `bundle install` and explain that to add a plugin you should add it as a gem as well as add config files

### DIFF
--- a/doc/installation_source.md
+++ b/doc/installation_source.md
@@ -8,6 +8,7 @@ Clone and install CLI core
 ```bash
 $ git clone https://github.com/theforeman/hammer-cli.git
 $ cd hammer-cli
+$ bundle install
 $ rake install
 $ cd ..
 ```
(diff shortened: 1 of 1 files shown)


doc/installation_source.md
```diff
@@ -25,7 +27,8 @@ and optionally other plugins.
 
 
 #### Step 2: enable and configure the plugins
-You'll have to copy configuration files to proper locations manually.
-Please check our [configuration instructions](installation.md#configuration)
+- Add the plugin as a gem to `Gemfile.local.rb`
+- Copt the configuration files to proper locations manually,
```
adamruzicka, line 31: Copt is a typo

MariaAga: It is, fixes, thanks!

ofedoren (approved):
Thanks, @MariaAga. Sorry for taking so long :)

PR #335, opened 2020-08-31
Fixes #30747 - Make fuzzy search work in hammer shell

ofedoren:
Needs to be CPed into `2.1-stable`


chris1984:
Starting to test

PR #339, opened 2021-01-16
Fixes #30996 - Better family assignment for options

ofedoren:
Now more options are being squeezed properly. Although not all of them if you install more plugins, but after some digging it seems that should be solved in those plugins.

The issue mentions more options than this patch will fix, but as you may notice they are related mostly to hammer-cli-katello and hammer-cli-foreman-discovery plugins. Those options I'll fix in the separate PRs to each of the plugins.

Currently this patch affects the following commands:
 -  user-group update: `--id` option fixed
 -  organization CRUD: `--organization-label` option returned
 
 For easier check of `--help` or `full-help` output differences you can use trick:
 Swap `!` here: https://github.com/theforeman/hammer-cli/blob/93056110a592403623bf72368c7ea98ab85e5a81/lib/hammer_cli/help/builder.rb#L38 and https://github.com/theforeman/hammer-cli/blob/93056110a592403623bf72368c7ea98ab85e5a81/lib/hammer_cli/help/builder.rb#L40

### DIFF
--- a/lib/hammer_cli/options/option_family.rb
+++ b/lib/hammer_cli/options/option_family.rb
@@ -5,7 +5,7 @@ module HammerCLI
     class OptionFamily
       attr_reader :children
 
-      IDS_REGEX = /\s?([Ii][Dd][s]?)\W|([Ii][Dd][s]?\Z)/
+      IDS_REGEX = /(\A[Ii][Dd][s]?)|\s([Ii][Dd][s]?)\W|([Ii][Dd][s]?\Z)/
 
       def initialize(options = {})
         @all = []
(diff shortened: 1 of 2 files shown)


lib/hammer_cli/options/option_family.rb
```diff
@@ -67,6 +74,9 @@ def child(switches, type, description, opts = {}, &block)
       end
 
       def adopt(child)
+        raise ArgumentError, 'Parent cannot be a child within the same family' if child == @parent
```
ofedoren, line 77: Not sure though if this should raise an exception (to prevent bugs) or just print into debug log (for devs info).

yifatmakias (approved):
Thanks @ofedoren. looks good to me.

PR #340, opened 2021-01-18
Fixes #31563 - Right usage in help output for aliases

ofedoren:
This patch fixes an issue that appears if someone tries to use a command in "fuzzy" mode to see its help: the output doesn't contain subcommand names.

To test this try to run e.g. `hammer arch l` or similar (not using full names) before and after this patch. Also should work in hammer shell.


yifatmakias (approved):

yifatmakias:
@ofedoren Works for me and looks good :)

shiramax (approved):
thanks @ofedoren !

PR #341, opened 2021-03-01
Fixes #29093 - Add full comparison for fields option

ofedoren:


lib/hammer_cli/output/definition.rb
```diff
@@ -73,7 +73,7 @@ def fields_data(fields, sets)
 
         data << sets.each_with_object({}) do |set, res|
           mark = field.sets.include?(set) ? 'x' : ''
-          value = set == sets.first ? field.full_label : mark
+          value = set == sets.first ? field.full_label.capitalize : mark
```
ofedoren, line 76: This is due to misunderstandings and misusage of the `--fields` option when there e.g. `ID` in help output, but should be used `Id` or `id` instead.

ofedoren:
@yifatmakias, now it's ready for review :)

yifatmakias (approved):
Thanks @ofedoren. Looks good to me.

PR #343, opened 2021-05-08
Fixes #32568 - Make cmd object be available in command extensions

ofedoren:
SSIA


yifatmakias (approved):
Thanks @ofedoren :)
Code looks good and this is a useful addition.

PR #345, opened 2021-05-24
Fixes #29093 - change from superficial copy to deep copy of fields

adiabramovitch:


yifatmakias (approved):
Looks good and works as expected :)

yifatmakias:
Thanks @adiabramovitch!

ekohl:
I think this caused a regression and has broken our nightlies. https://community.theforeman.org/t/foreman-nightly-rpm-pipeline-1068-failed/23813/2

jturel:
We're also seeing unit test failures in hammer-cli-katello

PR #348, opened 2021-07-13
Fixes #29093 - change from superficial copy to deep copy of fields

adiabramovitch:


yifatmakias (changes requested):
Thanks @adiabramovitch ,
Notice that the hammer_cli_foreman tests are failing with this error:
image

lib/hammer_cli/options/option_family.rb
```diff
@@ -11,7 +11,8 @@ def initialize(options = {})
         @all = []
         @children = []
         @options = options
-        @creator = options[:creator] || Class.new(HammerCLI::Apipie::Command)
```
ofedoren, line 14: FYI: I'm currently working on a little bit different approach for option creation and this anonymous class will disappear :) So, :+1: for the solution, but it should wait a bit before we get rid of this class.

ofedoren (approved):
Thanks, @adiabramovitch! Since this change doesn't break tests for plugins anymore we can finally merge it :)

ekohl:
Looks like this broke nightly again, just like in https://github.com/theforeman/hammer-cli/pull/345#issuecomment-852052250: https://ci.theforeman.org/job/foreman-nightly-rpm-pipeline/1160/tapResults/ 
```tap
not ok 6 import motd puppet class
# (in test file fb-test-puppet.bats, line 64)
#   `hammer proxy import-classes --name $(hostname -f)' failed with status 70
# Error: singleton can't be dumped
```

ofedoren:
> Looks like this broke nightly again, just like in [#345 (comment)](https://github.com/theforeman/hammer-cli/pull/345#issuecomment-852052250): https://ci.theforeman.org/job/foreman-nightly-rpm-pipeline/1160/tapResults/

@ekohl, yeap, I'm aware of that: https://github.com/theforeman/hammer-cli/pull/353

ekohl:
Thanks!

PR #349, opened 2021-07-25
Fixes #32783 - Change option details section and more

ofedoren:
This patch introduces a new section "Option details" with listed
option types and accepted values. Changes the way for automatic
option building making API params to be the main options. Also
deprecation warning will be shown only when a specific option is
actually being used.

PR #350, opened 2021-08-06
Fixes #32783 - Extract descs to option details section

ofedoren:
@yifatmakias, I've extracted this part from https://github.com/theforeman/hammer-cli/pull/349 and those changes are needed to only fix the described issue.

For easier review:

1. `Option details` section should be added now only if there more than one default option (`--help`).
2. `Option details` section now contains new entries describing what can be used as a value for an option.
3. Those entries are being generated automatically based on available option `Normalizer`s.

### DIFF
--- a/lib/hammer_cli/options/option_family.rb
+++ b/lib/hammer_cli/options/option_family.rb
@@ -5,7 +5,7 @@ module HammerCLI
     class OptionFamily
       attr_reader :children
 
-      IDS_REGEX = /(\A[Ii][Dd][s]?)|\s([Ii][Dd][s]?)\W|([Ii][Dd][s]?\Z)/
+      IDS_REGEX = /(\A[Ii][Dd][s]?)|\s([Ii][Dd][s]?)\W|([Ii][Dd][s]?\Z)|(numeric identifier|identifier)/.freeze
 
       def initialize(options = {})
         @all = []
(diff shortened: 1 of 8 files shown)


lib/hammer_cli/options/option_family.rb
```diff
@@ -33,6 +33,21 @@ def description
         end
       end
 
+      def help
+        [help_lhs, help_rhs]
```
yifatmakias, line 37: what is the meaning of help_lhs and help_rhs? Could you maybe change to more clear variable names?

yifatmakias (commented):
Thanks @ofedoren! over all looks good and also it works as it should. Just a small thing that I cloud not understand

lib/hammer_cli/options/option_family.rb
ofedoren: This is the same pattern as with Clamp:

https://github.com/mdub/clamp/blob/b95498a2c44c48d1703548dc20625ec528b97166/lib/clamp/attribute/definition.rb#L28-L30

https://github.com/mdub/clamp/blob/b95498a2c44c48d1703548dc20625ec528b97166/lib/clamp/option/definition.rb#L66

The meaning is `help_left_help_string` and `help_right_help_string`.

yifatmakias: oh, I understand now

yifatmakias (approved):
Thanks @ofedoren 
Code looks good, and it works as expected.

PR #351, opened 2021-08-06
Fixes #33225 - Show depr warning only on option usage

ofedoren:


yifatmakias (approved):
Thanks @ofedoren looks good :)

PR #353, opened 2021-08-16
Fixes #33259 - Don't store @context in field params

ofedoren:


tbrisker (approved):
Let's give it a try, thanks @ofedoren !

PR #355, opened 2021-11-02
Fixes #33810 - Remove a space in hammer's shebang

ekohl:
This space confuses Debian packaging on Ubuntu 20.04 and causes it to not be replaced. The end result is that it may end up using the wrong Ruby.


evgeni (approved):
With this change, Debian tooling correctly replaces the shebang with `#!/usr/bin/ruby` inside the package. Thanks!

PR #356, opened 2021-12-03
Fixes #34079 - Allow explicit strings in key=value options

ofedoren:
@adamruzicka, sorry for disturbing you with this, but since you reviewed the issue, could you please confirm this fixes the issue and doesn't bring new ones? :)

### DIFF
--- a/lib/hammer_cli/options/normalizers.rb
+++ b/lib/hammer_cli/options/normalizers.rb
@@ -104,15 +104,13 @@ module HammerCLI
 
         def strip_value(value)
           if value.is_a? Array
-            value.map do |item|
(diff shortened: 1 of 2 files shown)


adamruzicka (approved):
I played with it for a bit and I like it. I'd say this is the way it should be.

ofedoren:
@adiabramovitch, could you please check this one as well? :) Also, please let me know if you don't have merge rights (you should have, but recently there was a cleaning up in the teams of the organization, so I'm not quite sure if it's still true)

adiabramovitch:
@ofedoren Thanks Oleh! 
Just notice to use name="\"*\""

PR #359, opened 2022-02-16
Fixes #28283 - Autodetect if apidoc cache needs reload

ofedoren:
This is only a base to use in plugins, by default it's off (only a cache file is being created the first time). The usage is quite simple: https://github.com/theforeman/hammer-cli-foreman/pull/594.

This fix adds a new checksum based on installed plugins and their versions. If there was a new plugin installed or an old one was upgraded, then the old cache will be removed and thus a new cache will be downloaded automatically.

The only issue that persists is `hammer command action --help` won't show new options if the server was upgraded and added new params **unless** the corresponding hammer plugin was upgraded as well or `--reload-cache` was called manually or it's a brand new hammer installation. This is due to how `apipie-bindings` receives and uses stored cache. There is an option to force invalidating of the cache each time we run hammer, but it still needs to talk to the server (to load cache's checksum), but I don't find that a good idea since it's quite unnecessary to call server's API to simply show the help output (even if it will be fully updated).

I was thinking about other ways to detect when the autoreloading needs to be done, but since hammer can be installed as a simple gem and on a different machine than the server, there is no other way then to force an API call each time the hammer is being used (even for the help output only).

UPD: **Don't** merge just yet, I found out that it behaives differently in production...

### DIFF
--- a/lib/hammer_cli/modules.rb
+++ b/lib/hammer_cli/modules.rb
@@ -76,7 +76,8 @@ module HammerCLI
     end
 
     def self.load_all
-      HammerCLI::Modules.names.each do |m|
+      module_names = HammerCLI::Modules.names
+      module_names.each do |m|
         Modules.load(m)
       end
       loaded_for_deps = loaded_modules & disabled_modules
@@ -84,6 +85,36 @@ module HammerCLI
         message = _("Error: Some of the required modules are disabled in configuration: %s.") % loaded_for_deps.join(', ')
         raise HammerCLI::ModuleDisabledButRequired.new(message)
       end
+      return if modules_checksum
+
+      File.open(checksum_file_path, 'w') do |f|
+        f.write(compute_modules_checksum(module_names))
+      end
+    end
+
+    def self.module_version(mod)
+      Gem.loaded_specs[mod]&.version&.to_s
+    end
+
+    def self.compute_modules_checksum(names)
+      mods_with_version = names.map do |m|
+        "#{m}-#{module_version(m)}"
+      end.join(',')
+      Digest::SHA1.hexdigest(mods_with_version)
+    end
+
+    def self.modules_checksum
+      return unless File.exist?(checksum_file_path)
+
+      File.read(checksum_file_path)
+    end
+
+    def self.changed?
+      modules_checksum != compute_modules_checksum(HammerCLI::Modules.names)
+    end
+
+    def self.checksum_file_path
+      @checksum_file_path ||= File.expand_path(HammerCLI::Settings.get(:checksum_cache_file))
     end
 
     protected
@@ -95,7 +126,5 @@ module HammerCLI
     def self.logger
       Logging.logger['Modules']
     end
-
   end
-
 end
--- a/lib/hammer_cli/utils.rb
+++ b/lib/hammer_cli/utils.rb
@@ -67,8 +67,10 @@ module HammerCLI
   end
 
   def self.clear_cache
-    cache_file = File.expand_path(HammerCLI::Settings.get(:completion_cache_file))
-    File.delete(cache_file) if File.exist?(cache_file)
+    %i[checksum_cache_file completion_cache_file].each do |f|
+      cache_file = File.expand_path(HammerCLI::Settings.get(f))
+      File.delete(cache_file) if File.exist?(cache_file)
+    end
   end
 
   def self.interactive?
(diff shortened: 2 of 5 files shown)


lib/hammer_cli/modules.rb
```diff
@@ -76,14 +76,45 @@ def self.require_module(name)
...
       loaded_for_deps = loaded_modules & disabled_modules
       unless loaded_for_deps.empty?
         message = _("Error: Some of the required modules are disabled in configuration: %s.") % loaded_for_deps.join(', ')
         raise HammerCLI::ModuleDisabledButRequired.new(message)
       end
+      return if modules_checksum
+
+      File.open(checksum_file_path, 'w') do |f|
+        f.write(compute_modules_checksum(module_names))
+      end
```
adamruzicka, line 92: There's a shorthand for open-write-close
```suggestion
      File.write(checksum_file_path, compute_modules_checksum(module_names))
```

lib/hammer_cli/modules.rb
```diff
@@ -76,14 +76,45 @@ def self.require_module(name)
...
+
+      File.read(checksum_file_path)
+    end
+
+    def self.changed?
+      modules_checksum != compute_modules_checksum(HammerCLI::Modules.names)
+    end
+
+    def self.checksum_file_path
+      @checksum_file_path ||= File.expand_path(HammerCLI::Settings.get(:checksum_cache_file))
```
adamruzicka, line 117: Can this setting be unset?

lib/hammer_cli/utils.rb
```diff
@@ -67,8 +67,10 @@ def self.tty?
   end
 
   def self.clear_cache
-    cache_file = File.expand_path(HammerCLI::Settings.get(:completion_cache_file))
-    File.delete(cache_file) if File.exist?(cache_file)
+    %i[checksum_cache_file completion_cache_file].each do |f|
+      cache_file = File.expand_path(HammerCLI::Settings.get(f))
```
adamruzicka, line 71: Can these settings be unset?

adamruzicka (commented):
> Don't merge just yet, I found out that it behaives differently in production

How does it behave in production?

> but it still needs to talk to the server (to load cache's checksum), but I don't find that a good idea since it's quite unnecessary to call server's API

While that is true, it is the only reliable way of being always up-to-date. What if the server honored the `If-Modified-Since` header (and hammer was sending it) to minimize the amount of data that needs to be gathered on the server side and transferred? If checking if local cache is fresh would be cheap enough then maybe hitting the server every time would be bearable?

ofedoren:
> How does it behave in production?

When I was writing this I forgot that `$LOAD_PATH` in my dev env contains all the plugins automatically and thus some constants/specs are available even without explicit `require`. `Gem.loaded_specs` had all the plugins listed even before we `require` them in `Modules.load_all` and thus in this [line](https://github.com/theforeman/hammer-cli/blob/5a493c0a28a2ecefb557443236e2699a661a6611/lib/hammer_cli/apipie/api_connection.rb#L11) `Modules.changed?` computed the right cheksum of plugins with their versions. But in production the `Gem.loaded_specs` is updated only after explicit `require` we do in `Modules.load_all`. The problem is that `Modules.changed?` is called before we load all the plugins to check their version (compute checksum) and thus it's being a wrong one, and due to that `Modules.changed?` always `true` since we save the right hash (after all is loaded), but comparing against wrong one (built by `Modules.changed?`).

The problem is that we need to have all the plugins loaded to compute the right checksum before we try to load the cache, but at the same time, to load (`require`) all the plugins, we need the cache first :/

> While that is true, it is the only reliable way of being always up-to-date. What if the server honored the `If-Modified-Since` header (and hammer was sending it) to minimize the amount of data that needs to be gathered on the server side and transferred? If checking if local cache is fresh would be cheap enough then maybe hitting the server every time would be bearable?

Hmm, didn't consider that, but since I'm quite stuck, I'll try to check this out.

lib/hammer_cli/modules.rb
ofedoren: Yes, but the supported way is to simply comment the setting out in the config file. In this case defaults will be used, so the value of this setting will be always present.

The only way to break this is to set `:checksum_cache_file: nil` in the config file, but AFAIK this never happened before, otherwise we would have updated the way we treat unsetting cache file paths.

ofedoren:
Closed in favor https://github.com/theforeman/hammer-cli-foreman/pull/599.

PR #360, opened 2022-04-05
Fixes #34724 - Fix hammer shell with pipe as stdin

adamruzicka:
Hammer shell tries to store terminal properties with `stty -g` and restore
them once it is done. This works rather poorly when the input is not a
tty (a pipe). When run in such a way, hammer complains about a couple of
things, depending on platform, possibly caused by different `stty`
versions.

This should prevent the following errors from showing up:
```
stty: 'standard input': Inappropriate ioctl for device
```
```
stty: invalid argument ‘’
Try 'stty --help' for more information.
```


ofedoren (approved):
Thanks, @adamruzicka, works as expected :)

PR #362, opened 2023-01-04
Automatically pull in languages with > 50% translated

ekohl:
Also takes a stab at making the locale generic, but it isn't quite there.

### DIFF
--- a/locale/Makefile
+++ b/locale/Makefile
@@ -4,7 +4,7 @@
 # Run make help to list available targets
 #
 DOMAIN = hammer-cli
-VERSION = $(shell bundle exec ruby -e 'require "rubygems"; spec = Gem::Specification::load("../hammer_cli.gemspec"); puts spec.version')
-MAIN_MAKEFILE = $(shell bundle exec ruby -e 'require "hammer_cli"; puts HammerCLI::I18n.main_makefile')
+VERSION = $(shell ruby -rrubygems -e 'puts Gem::Specification::load(Dir.glob("../*.gemspec").first).version')
+MAIN_MAKEFILE = $(shell bundle exec ruby -rhammer_cli -e 'puts HammerCLI::I18n.main_makefile')
 
 include $(MAIN_MAKEFILE)
--- a/locale/Makefile.def
+++ b/locale/Makefile.def
@@ -40,9 +40,14 @@ uniq-po: ## Unify duplicate translations in .po files
 %.edit.po:
 	touch $@
 
-# Pull translaions from transifex
+# Pull in all languages with at least 50%
+.PHONY: tx-new
+tx-new:
+	tx pull --minimum-perc 50 --all
+
+# Pull translations from transifex
 .PHONY: tx-pull
-tx-pull: $(EDITFILES)
+tx-pull: tx-new $(EDITFILES)
 	@command -v tx >/dev/null 2>&1 || { echo "Command tx not found. Make sure you have transifex-client installed and configured." >&2; exit 1; }
 	tx pull -f
 	for f in $(EDITFILES); do \
--- a/locale/README.md
+++ b/locale/README.md
@@ -1,15 +1,9 @@
 Updating the translations
 -------------------------
 
-  1. Check if there are any new languages with progress more than 50% on [transifex](https://www.transifex.com/projects/p/foreman/resource/hammer-cli/). If so, do the following for each of the new languages:
+  1. Make sure you have `transifex-client` installed
 
-  ```
-    mkdir locale/<lang>
-    cp locale/hammer-cli.pot locale/<lang>/hammer-cli.po
-  ```
-  2. Make sure you have `transifex-client` installed
-
-  3. Update the translations. From GIT repo root directory run:
+  2. Update the translations. From GIT repo root directory run:
 
   ```
     make -C locale tx-update


locale/Makefile
```diff
@@ -4,7 +4,7 @@
 # Run make help to list available targets
 #
 DOMAIN = hammer-cli
```
ekohl, line 6: I thought about extracting this from the gemspec too. For example, you could make some logic in Hammer to print the domain for a given gem. If it doesn't match the gem name, it could use the gem metadata.

For reference, we have [this](https://github.com/theforeman/foreman/blob/acafe1aae2eef439a73334635046dfb79be55894/script/plugin_webpack_directories.rb#L44) in Foreman. We could introduce a metadata key like i18n_domain.

Then the shared gemspec becomes 2 lines: gem name and `MAIN_MAKEFILE`.

@ofedoren any thoughts on this?

ekohl (commented):
@ofedoren I'm currently playing with this to make it easy to set up translations everywhere, taking the original idea further.

locale/Makefile
```diff
@@ -4,7 +4,7 @@
...
-VERSION = $(shell bundle exec ruby -e 'require "rubygems"; spec = Gem::Specification::load("../hammer_cli.gemspec"); puts spec.version')
-MAIN_MAKEFILE = $(shell bundle exec ruby -e 'require "hammer_cli"; puts HammerCLI::I18n.main_makefile')
+VERSION = $(shell ruby -rrubygems -e 'puts Gem::Specification::load(Dir.glob("../*.gemspec").first).version')
+MAIN_MAKEFILE = $(shell bundle exec ruby -rhammer_cli -e 'puts HammerCLI::I18n.main_makefile')
```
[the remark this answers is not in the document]
ofedoren: I guess I know the difference: in the old version `require "hammer_cli"` is executed in context of `bundle exec` loading local gem, but in the new version `-rhammer_cli` is executed in context of `ruby` ignoring `bundle exec` and since there is no installed `hammer_cli` gem it fails.

It does work for `VERSION` since `rubygems` is available globally.

locale/Makefile
ofedoren: I guess we could go further:
 - Move `VERSION` and `DOMAIN` into `Makefile.def` since it will be shared and used in plugins. No need to redefine those each time. That way, the only thing a plugin will need is a `Makefile` with two lines: importing shared makefile and including it.
 - Define `VERSION` as per your suggestion.
 - Define `metadata` as per your suggestion in gemspec like
 ```ruby
 require 'hammer_cli/i18n'
 # ...
 s.metadata      = { 'in_domain' => HammerCLI::I18n::LocaleDomain.new.domain_name }
 ```
  - Define `DOMAIN` as 
 ```Makefile
 DOMAIN = $(shell ruby -rrubygems -e '
spec = Gem::Specification::load(Dir.glob("../*.gemspec").first);
in_domain = spec.respond_to?(:metadata) && spec.metadata["in_domain"]
puts in_domain || spec.name')
 ```
Pros: we have automated domain definition and only one source of truth. Cons: we need to update all plugins to be consistent since if a plugin won't have 'metadata' entry then the only way we can guess is to use gem's name, which may differ from the domain.

locale/Makefile.def
```diff
@@ -40,9 +40,14 @@ uniq-po: ## Unify duplicate translations in .po files
...
+.PHONY: tx-new
+tx-new:
+	tx pull --minimum-perc 50 --all
+
+# Pull translations from transifex
 .PHONY: tx-pull
-tx-pull: $(EDITFILES)
+tx-pull: tx-new $(EDITFILES)
 	@command -v tx >/dev/null 2>&1 || { echo "Command tx not found. Make sure you have transifex-client installed and configured." >&2; exit 1; }
 	tx pull -f
```
ofedoren, line 52:
```suggestion
	tx pull -f --minimum-perc 50 --all
```

ofedoren (commented):
Thanks, @ekohl, so sorry I just took a look now. Few suggestions:

locale/Makefile
ofedoren: Sounds good, the only difference then will be something like `DOMAIN = $(DOMAIN) || $(shell ...)`.

locale/Makefile.def
```diff
@@ -40,9 +40,14 @@ uniq-po: ## Unify duplicate translations in .po files
 %.edit.po:
 	touch $@
 
-# Pull translaions from transifex
+# Pull in all languages with at least 50%
+.PHONY: tx-new
+tx-new:
+	tx pull --minimum-perc 50 --all
```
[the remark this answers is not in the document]
ekohl: My reasoning was that I wanted to separate the "Add new languages" step from "Update languages". For example, I don't know if `tx pull -f --minimum-perc 50 --all` only pulls translations that are >= 50%.

Essentially I'm thinking of the scenario where you have:
* Language XX is added because it's >= 50%
* Later language XX drops below 50%, but there is a new string

In that case you could argue XX should be dropped, but my current intention is that you at least pull in the new string. While I didn't test it, I thought that a single command would ignore the string.

That said, I think `tx-new` now also needs some step to create the `%{DOMAIN}.po` files after pulling so this step may be expanded anyway.

ofedoren:
Closing in favor of https://github.com/theforeman/hammer-cli/pull/366

PR #365, opened 2023-04-07
Config template typo

wbclark:
Something I noticed while looking into adding a hammer source install to katello dev environment


config/cli_config.template.yml
```diff
@@ -17,7 +17,7 @@
 # Enable/disable color output of logger in Clamp commands
 :watch_plain: false
 
-# Forece relaod of Apipie cache with every Hammer invocation
+# Force relaod of Apipie cache with every Hammer invocation
```
ofedoren, line 20: Since you're at it, there is another typo just next to this one :)
```suggestion
# Force reload of Apipie cache with every Hammer invocation
```

ofedoren (approved):
Thanks, @wbclark :)

PR #367, opened 2023-04-27
Fixes #36337 - Update rake and clamp deps

ofedoren:
This PR updates `rake` to any latest available and updates `clamp` to be any version, but `2.0.0` (currently, the latest is 1.3.2).

There is also some refactoring due to `clamp` changes and some refactoring due to `rake` spamming a lot of Ruby warnings. I fixed majority of them (mostly it was `uninitialized variable` warning, although I'm not sure that's a good idea: https://bugs.ruby-lang.org/issues/17055#note-6)

Anyway, there is a new `warning` flag for `rake`'s `TestTask`, which suppresses all Ruby warnings. Without this flag quite a lot of our tests will fail since they depend on strict output. I left `RUBY_WARNINGS` env variable to be set to run the tests with Ruby warnings.

### DIFF
--- a/hammer_cli.gemspec
+++ b/hammer_cli.gemspec
@@ -27,7 +27,9 @@ EOF
   s.require_paths = ["lib"]
   s.executables = ['hammer', 'hammer-complete']
 
-  s.add_dependency 'clamp', '>= 1.1', '< 1.2.0'
+  s.required_ruby_version = '>= 2.7'
+
+  s.add_dependency 'clamp', '>= 1.3.1', '< 2.0.0'
   s.add_dependency 'logging'
   s.add_dependency 'unicode-display_width'
   s.add_dependency 'unicode'
--- a/lib/hammer_cli/abstract.rb
+++ b/lib/hammer_cli/abstract.rb
@@ -192,7 +192,7 @@ module HammerCLI
 
     def self.output(definition=nil, &block)
       dsl = HammerCLI::Output::Dsl.new
-      dsl.build &block if block_given?
+      dsl.build(&block) if block_given?
       output_definition.append definition.fields unless definition.nil?
       output_definition.append dsl.fields
     end
--- a/lib/hammer_cli/logger.rb
+++ b/lib/hammer_cli/logger.rb
@@ -54,7 +54,7 @@ module HammerCLI
       log_dir = File.expand_path(HammerCLI::Settings.get(:log_dir) || DEFAULT_LOG_DIR)
       begin
         FileUtils.mkdir_p(log_dir, :mode => 0750)
-      rescue Errno::EACCES => e
+      rescue Errno::EACCES
         $stderr.puts _("No permissions to create log dir %s.") % log_dir
       end
 
@@ -68,7 +68,7 @@ module HammerCLI
                                                             :size     => (HammerCLI::Settings.get(:log_size) || 1)*1024*1024) # 1MB
         # set owner and group (it's ignored if attribute is nil)
         FileUtils.chown HammerCLI::Settings.get(:log_owner), HammerCLI::Settings.get(:log_group), filename
-      rescue ArgumentError => e
+      rescue ArgumentError
         $stderr.puts _("File %s not writeable, won't log anything to the file!") % filename
       end
 
--- a/lib/hammer_cli/subcommand.rb
+++ b/lib/hammer_cli/subcommand.rb
@@ -11,6 +11,7 @@ module HammerCLI
         @subcommand_class = subcommand_class
         @hidden = options[:hidden]
         @warning = options[:warning]
+        super(@names, @description, @subcommand_class)
       end
 
       def hidden?
--- a/lib/hammer_cli/utils.rb
+++ b/lib/hammer_cli/utils.rb
@@ -48,18 +48,6 @@ class String
   end
 end
 
-class Hash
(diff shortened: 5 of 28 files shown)


ofedoren:
[test hammer]

lib/hammer_cli/logger.rb
```diff
@@ -54,7 +54,7 @@ def self.initialize_logger(logger)
       log_dir = File.expand_path(HammerCLI::Settings.get(:log_dir) || DEFAULT_LOG_DIR)
       begin
         FileUtils.mkdir_p(log_dir, :mode => 0750)
-      rescue Errno::EACCES => e
+      rescue Errno::EACCES => _
```
ekohl, line 57: Repeated pattern, so only commenting once.
```suggestion
      rescue Errno::EACCES
```

ofedoren:
Thanks, @ekohl, updated.

lib/hammer_cli/abstract.rb
```diff
@@ -341,27 +341,27 @@ def exception_handler_class
     end
 
     def self.desc(desc=nil)
-      @desc = desc if desc
-      @desc
+      return @desc ||= nil unless desc
```
ekohl, line 344: I think this is a lot harder to read. It's also a bit of an odd API where it's both a setter and a getter.

Can you separate the undefined variable changes to a separate commit? AFAIK they're warnings, not errors so it should be possible to update rake without addressing them immediately.

ekohl (changes requested):
I think the undefined variables changes make the code harder to understand. It'd be good to separate those out and review them separately.

ofedoren:
Thanks, @ekohl, I've splitted all the changes here to separate commits. I hope this will make review easier.

Also, since https://github.com/theforeman/hammer-cli-foreman/pull/613 contains quite similar changes, I'll update that one after this PR gets resolved/approved.

hammer_cli.gemspec
```diff
@@ -27,7 +27,7 @@ EOF
   s.require_paths = ["lib"]
   s.executables = ['hammer', 'hammer-complete']
 
-  s.add_dependency 'clamp', '>= 1.1', '< 1.2.0'
+  s.add_dependency 'clamp', '>= 1.1', '< 2.0.0'
```
ekohl, line 30: You use the `line` method introduced in clamp 1.3.1 so that should be the lower bound. We'll need to deal with that in packaging, but I think that's ok.

lib/hammer_cli/utils.rb
```diff
@@ -48,18 +48,6 @@ def wrap(line_width: 80, break_sequence: "\n")
   end
 end
 
-class Hash
-  # for ruby < 2.5.0
```
ekohl, line 52: Does it make sense to set `required_ruby_version` to `>= 2.5` in the gemspec?

lib/hammer_cli/abstract.rb
```diff
@@ -341,27 +341,27 @@ def exception_handler_class
...
     end
 
     def self.command_names(*names)
-      @names = names unless names.empty?
-      @names || (superclass.respond_to?(:command_names) ? superclass.command_names : nil)
+      names = names.compact
+      return @names ||= (superclass.respond_to?(:command_names) ? superclass.command_names : nil) if names.empty?
+
+      @names ||= []
+      @names |= names
```
ekohl, line 358: This changes the implementation and now caches the response of the superclass. Is that intentional? I'd keep the change minimal:
```suggestion
      @names ||= nil
      @names = names unless names.empty?
      @names || (superclass.respond_to?(:command_names) ? superclass.command_names : nil)

**Review by ekohl — COMMENTED** [review]
https://stackoverflow.com/questions/2098742/ruby-instance-variable-not-initialized-warning/2098835#2098835 suggests the undefined warning goes away with Ruby 3.0. Since you disable warnings anyway, perhaps the last commit should be trimmed down or even skipped.

File: lib/hammer_cli/utils.rb (hunk 4 above)
Comment (ofedoren) [reply lib/hammer_cli/utils.rb:52]: Well, I'd set it to `>= 2.7` even, since I'm not sure we even test lower versions...

**Comment by ofedoren** [comment]
> suggests the undefined warning goes away with Ruby 3.0

And I found the commit with that :) https://github.com/ruby/ruby/commit/01b7d5acc702df22d306ae95f1a9c3096e63e624

So, I've decided to drop the last commit. Hammer is already slow, it's not worth to slow it more.

Updated, but without new changes related to uninitialized variables.

File: lib/hammer_cli/subcommand.rb [hunk 6]
```diff
@@ -11,6 +11,7 @@ def initialize(names, description, subcommand_class, options = {})
         @subcommand_class = subcommand_class
         @hidden = options[:hidden]
         @warning = options[:warning]
+        super(@names, @description, @subcommand_class)
```
ekohl, line 14: If you call this, I wonder if names, description and subclass_command aren't already stored. I think you can drop `@names =` etc.

ekohl (approved):

lib/hammer_cli/subcommand.rb
ofedoren: Not really, AFAIR I've added this line because something wasn't initialized in clamp itself, crashing the commands.

Also, I rather call `super` with explicit arguments because implicitly it would send `params` as well, which is not supported by clamp.

PR #368, opened 2023-05-10
Refs #36349 - Update minitest version

ofedoren:
Follows https://github.com/theforeman/hammer-cli-foreman/pull/611.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ group :test do
   gem 'minitest-spec-context'
   gem 'simplecov'
   gem 'mocha'
-  gem 'ci_reporter', '>= 1.6.3', "< 2.0.0", :require => false
+  gem 'ci_reporter_minitest', '~> 1.0', :require => false
 end
 
 # load local gemfile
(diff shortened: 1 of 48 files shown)


ofedoren:
Damn, I hoped that it's something on my machine causing this tests to fail...

Well, I've tried to find what caused the tests to fail **randomly**, but failed. I'll leave it here for now.

ehelms:
At a quick glance, the failing tests are all array comparisons. I've always had to solve those by adding `.sort` to ensure that the array has the same order every time.

ofedoren:
I thought about that, but it should not happen at all :/ This is somehow related to the newest minitest version, but I couldn't figure out what could cause the tests fail randomly. The only thing I could think off is there might be a change in code loading or test execution, but again, I couldn't find where/when this happened.

It's also possible we had broken tests all that time, but I doubt it (manually everything works as per tests definitions).

ofedoren:
Closing in favor of https://github.com/theforeman/hammer-cli/pull/372.

PR #374, opened 2023-07-27
Fixes #36258 - Improve regex for KeyValueList normalizer

ofedoren:
This fixes the issue, but as a downside, we will slightly change the old behavior: previously the normalizer will fail on inputs like `a=1,2,c=3`, which implies that `2` should've had `b` as a key. But we can't really decide for users whether it's a typo or the actual value, thus it will be treated as a value now, which is kinda make sense for quite simple values which don't work now :/

Other approach will be forcing comma escaping if the comma must be a part of the value, but this seems more safer in terms backwards compatibility (previously on wrong input the script will fail, but with that change the stuff that worked might stop working).

### DIFF
--- a/lib/hammer_cli/options/normalizers.rb
+++ b/lib/hammer_cli/options/normalizers.rb
@@ -50,9 +50,7 @@ module HammerCLI
       end
 
       class KeyValueList < AbstractNormalizer
-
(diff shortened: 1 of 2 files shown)


adamruzicka (approved):
The regexp is slightly bigger than what I'm comfortable with, but it solves the issue we were having.

adamruzicka:
Thank you @ofedoren !

PR #376, opened 2023-08-01
Drop unicode gem

ehelms:
As of Ruby 2.4 strings can natively handle unicode capitalization and the (very old) unicode gem is no longer required.


ofedoren (approved):
Thanks, @ehelms!

PR #366, opened 2023-04-18
Fixes #36283 - Update transifex automation

ofedoren:
Replaces https://github.com/theforeman/hammer-cli/pull/362.

@ekohl, I think you'd want to take a look at this. Eventually this is the same as yours https://github.com/theforeman/hammer-cli/pull/362, but with extra steps.

I've decided not to rely on `--minimum-perc 50`, but rather write a script to check lang percentage via API call (anyway it will be used rather once in a while) . If there are new langs with > 50% translated strings, it will create needed dir/files, so `tx pull` can proceed as usual.

Although, all plugins still need to be updated. Mostly due to `.tx/config` [1]. Without this change it seemed that I couldn't pull the translations at all :/ I'm going to do it anyway [2] (I'll populate the tracker a bit later), so we have consistency and also new strings pulled.

This PR also includes new pulled string as a separate commit. Notice, there are two new languages (cs_CZ and ka).

There is an example PR that will be similar to other plugins: https://github.com/theforeman/hammer-cli-foreman/pull/610

[1] - https://developers.transifex.com/docs/migrating-from-older-versions-of-the-client
[2] - https://projects.theforeman.org/issues/36314
[3] - https://github.com/theforeman/hammer-cli-foreman/blob/master/Rakefile#L15

UPD: Forgot to mention that after a plugin has translation support (domain registered, confg updated, Rakefile entry [3]) it needs to only run `make -C locale tx-update` within the checkout directory to pull the translations. Didn't test on production like setup though...

PR #377, opened 2023-09-26
Make update-po target working on Ubuntu

stejskalleos:


stejskalleos:
It's for translation github action: https://github.com/stejskalleos/hammer-cli-foreman-google/blob/akcion/.github/workflows/translations_pull.yml

ofedoren:
After https://github.com/theforeman/hammer-cli/pull/366 it's better to use rake task now: `bundle exec rake tx:update`. Would you mind to update https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/i18n/find_task.rb#L69 as well? Or have you tried the new command on Ubuntu?

stejskalleos:
Huh I didn't know that there is a rake task for it. So what's the correct way of updating strings in hammer plugins? Is it documented somewhere?

stejskalleos:
The rake task fails on Ubuntu with:
```
locale/hammer_cli_foreman_google.po" for reading: No such file or directory
rake aborted!
Command failed with status (1): [msgcat --use-first --no-location /root/ham...]
/root/hammer-cli/lib/hammer_cli/i18n/find_task.rb:69:in `block (3 levels) in define'
/root/hammer-cli/lib/hammer_cli/i18n/find_task.rb:67:in `each'
/root/hammer-cli/lib/hammer_cli/i18n/find_task.rb:67:in `block (2 levels) in define'
/root/hammer-cli/lib/hammer_cli/i18n/find_task.rb:88:in `block (2 levels) in define'
/root/.rbenv/versions/2.7.8/bin/bundle:23:in `load'
/root/.rbenv/versions/2.7.8/bin/bundle:23:in `<main>'
```

ofedoren:
@stejskalleos, I've just checked the rake task on Ubuntu and it just works™. Have you `sudo apt install gettext`?

stejskalleos:
```
apt install gettext

gettext is already the newest version (0.21-11).
```
On Ubuntu 23.04 it fails with the error in the comment above. `make tx-pull` works fine

ofedoren:
Well, it seems that makefile fails because it expects `bash` instead of `dash`, but the rake task should work since we don't use any substitution, just full path to the files...

I've checked on Ubuntu 20.04, but I'll try to re-check on Ubuntu 23.04...

ofedoren:
@stejskalleos, I've re-checked on Ubuntu 23.04 (just to make sure) and it still works:
ScreenShot-1696246087329
The only possible differences may be in what I have installed on the OS:
 - `sudo apt-get update && sudo apt install ruby-dev build-essential gettext ruby-bundler`
Also I use the new `tx` tool instead of the old one:
 - `curl -o- https://raw.githubusercontent.com/transifex/cli/master/install.sh | bash` (https://developers.transifex.com/docs/cli#installing-with-a-script-linuxmac)

stejskalleos:
Hmm, so it's something on my setup, closing the PR

PR #384, opened 2024-06-20
Fixes #37579 - Define correct timezone behavior in unittest

dosas:
see also https://github.com/theforeman/hammer-cli/pull/369#issuecomment-2180564378

### DIFF
--- a/test/unit/options/normalizers_test.rb
+++ b/test/unit/options/normalizers_test.rb
@@ -416,6 +416,16 @@ describe HammerCLI::Options::Normalizers do
 
   end
 
+  def mock_env(partial_env_hash)
+    old_env = ENV.to_hash
+    ENV.update partial_env_hash
+    begin
+      yield
+    ensure
+      ENV.replace old_env
+    end
+  end
+
   describe 'datetime' do
 
     let(:formatter) { HammerCLI::Options::Normalizers::DateTime.new }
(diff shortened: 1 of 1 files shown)


test/unit/options/normalizers_test.rb
```diff
@@ -429,15 +439,45 @@
...
+    it "should parse iso8601 compliant timezones" do
+      mock_env("TZ" => "UTC") do
+        _(formatter.format("1986-01-01 08:30:20 +04:00")).must_equal("1986-01-01T08:30:20+04:00")
+      end
+    end
+
+    it "should not recognize invalid timezones" do
+      mock_env("TZ" => "UTC") do
+        _(formatter.format("1986-01-01 08:30:20 bullshit")).must_equal("1986-01-01T08:30:20+00:00")
+      end
```
adamruzicka, line 468: Could you either duplicate this one or change it from utc to something else to be extra clear it's picking up the timezone from the environment and not silently defaulting to utc?

test/unit/options/normalizers_test.rb
```diff
@@ -429,15 +439,45 @@
...
+    end
+
+    it "should parse iso8601 compliant timezones" do
+      mock_env("TZ" => "UTC") do
+        _(formatter.format("1986-01-01 08:30:20 +04:00")).must_equal("1986-01-01T08:30:20+04:00")
+      end
+    end
+
+    it "should not recognize invalid timezones" do
```
adamruzicka, line 465: The name is slightly misleading as it also ignores "valid" timezones such as `EDT`.

test/unit/options/normalizers_test.rb
```diff
@@ -429,15 +439,45 @@
...
+      mock_env("TZ" => "UTC") do
+        _(formatter.format("1986-01-01 08:30:20 bullshit")).must_equal("1986-01-01T08:30:20+00:00")
```
adamruzicka, line 467: Although I personally don't mind, some might find this invalid time zone inappropriate

adamruzicka (changes requested):

test/unit/options/normalizers_test.rb
dosas: What name would you suggest? `should not recognize certain timezones` because it recognises some.

adamruzicka: maybe something along the lines of `should not recognize timezone strings not formatted according to iso8601`?

dosas: That is also slightly misleading because it recognizes timezones like `EST`

adamruzicka (approved):

adamruzicka:
Thank you @dosas !

PR #385, opened 2024-07-01
Log module loading warning to stderr

ekohl:
When requesting output as JSON the warnings on stdout will prevent loading. It also fixes a typo in occurred.


ofedoren:
The test failures are related :/ Nothing big, tests don't expect the warning in stderr, but stdout.

ekohl:
I think the current failure is unrelated. I can't reproduce it locally, but I opened https://github.com/theforeman/hammer-cli/pull/386 for it.

ofedoren (approved):
Thanks, @ekohl, the failure is indeed not related. I've merged the second PR, so after merge this should be green as well.

PR #386, opened 2024-07-02
Make clear_cache safe if no completion file is set

ekohl:
If it returns `nil`, `expand_path` fails. It also replaces a file deletion race condition with an atomic remove.

Observed a failure in https://github.com/theforeman/hammer-cli/pull/385. I suspect it failed because `HOME` wasn't set, but didn't dig in further.


ofedoren (approved):
Thanks, @ekohl, even though completion cache file should have a default value, this makes things safer anyway.

PR #387, opened 2024-11-12
pin FFI to < 1.17 for Ruby 2.7

evgeni:
FFI 1.17 needs rubygems 3.3.22+, which is Ruby 3.0+ only

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,9 @@ gemspec
 
 gem 'gettext', '>= 3.1.3', '< 4.0.0'
 
+# FFI 1.17 needs rubygems 3.3.22+, which is Ruby 3.0+ only
+gem "ffi", "<1.17" if RUBY_VERSION < '3.0'
+
 group :test do
   gem 'rake'
   gem 'thor'


Gemfile
```diff
@@ -4,6 +4,9 @@ gemspec
 
 gem 'gettext', '>= 3.1.3', '< 4.0.0'
 
+# FFI 1.17 needs rubygems 3.3.22+, which is Ruby 3.0+ only
+gem "ffi", "<1.17" if RUBY_VERSION < '3.0'
```
ofedoren, line 8: Shouldn't we rather use:

```suggestion
gem "ffi", "<1.17" if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('3.0')
```

?

ofedoren: Nvm, I think it's not available during Gemfile evaluation?..

evgeni: Technically, maybe.
But we've used the exact same notation in other places (https://github.com/theforeman/foreman/blob/develop/Gemfile#L55) and it works fine.

ofedoren (approved):
Thanks, @evgeni, if it works fine in Foreman then fine by me :)

PR #390, opened 2024-12-11
Fixes #38077 - Don't build hidden params as options

ofedoren:
Requires:
 - https://github.com/Apipie/apipie-bindings/pull/84

Used in:
 - https://github.com/theforeman/hammer-cli-foreman/pull/634

@adamlazik1, you might be interested in this, since it tries to use your changes in https://github.com/theforeman/foreman/pull/10322.


ofedoren:
We'd probably want to merge https://github.com/theforeman/hammer-cli/pull/391 first, so we can add a commit here to check if the tests are green with merged changes to apipie-bindings.

adamlazik1 (approved):
LGTM

PR #394, opened 2025-02-10
Ruby 3.4 compatibility

ekohl:
This is a warning raised on Ruby 3.3:

    lib/hammer_cli/output/adapter.rb:5: warning: csv was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.4.0.

It also raises one about syslog:

    lib/hammer_cli/settings.rb:2: warning: syslog was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.4.0.

However, this one is pulled in via logging and we don't care about syslog. https://github.com/TwP/logging/issues/247 says this is a Ruby bug so we may need to accept it.

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -20,13 +20,13 @@ jobs:
 
     steps:
       - name: Get apipie-bindings
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           repository: Apipie/apipie-bindings
           ref: master
           path: apipie-bindings
       - name: Get hammer-cli
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           path: hammer-cli
       - name: Configure local gem dependencies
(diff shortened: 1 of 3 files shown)


ekohl:
Looks like it fails because `oauth` doesn't properly depend on `base64`. @ofedoren what do you think we should do? Try to get oauth fixed or depend on it here with a comment why?

ofedoren:
@ekohl, for simplicity I'd go with explicit dependencies with a comment to check if it gets fixed once we're dropping older Ruby's.

ekohl:
Looks like there's a difference in the `Hash#inspect` output that makes this test fail:
https://github.com/theforeman/hammer-cli/blob/a8e710f4e6dab0527b03d0d3ed880d4403666e80/test/unit/abstract_test.rb#L188

The line in question is here:
https://github.com/theforeman/hammer-cli/blob/a8e710f4e6dab0527b03d0d3ed880d4403666e80/lib/hammer_cli/abstract.rb#L117

I'll leave this until after Foreman 3.14 branching but feel free to pick it up if you want.

ofedoren (approved):
Thanks, @ekohl, let's get this in :)

PR #395, opened 2025-05-07
Fixes #38410 - Add mechanism for deprecating displayed fields

qcjames53:
Currently, hammer has no way to deprecate old fields our developers wish to remove. This PR adds an option for deprecating a field with `:deprecated => true`. When a deprecated field is displayed, a stderr warning message is shown indicating it will be removed in the future. Updated `doc/creating_commands.md` to show how to deprecate a field.

**How to test**:

1. Find a field in hammer-cli-foreman or hammer-cli-katello and modify it to have the deprecated flag:
```
field :example, _("Example field"), Fields::Field, :deprecated => true
```
2. When displaying this field, verify that a warning is printed to stderr:
```
WARNING: Field 'Example field' is deprecated and may be removed in future versions.
```
3. Remove this field from the default set and confirm that the error message does not display when the field is not rendered:
```
field :example, _("Example field"), Fields::Field, :sets => ['ALL'], :deprecated => true
```
4. Confirm that the error message does display for a variety of field output edge cases (`--fields`, etc). Ensure output is always sent to stderr.
5. Add the :replaced_by flag and repeat the previous testing steps:
```
field :example, _("Example field"), Fields::Field, :replaced_by => "Example/Path"
```

### DIFF
--- a/doc/creating_commands.md
+++ b/doc/creating_commands.md
@@ -755,6 +755,35 @@ def adapter
 end
 ```
 
+#### Deprecating fields
+To deprecate a field, add `:deprecated => true` as an option for the field. This will print a warning message to stderr whenever the field is displayed. Consider removing this field from the default set so it is not displayed without a `--fields` param:
+
+```
+field :dep_fld, _("Deprecated field"), Fields::Field, :sets => ['ALL'], :deprecated => true
+```
+
+Example output:
+
+```
+$ hammer foo info --fields "Deprecated field"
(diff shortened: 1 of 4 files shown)


qcjames53:
@ofedoren Thank you for the review! I have added your requested changes and fixed the commit name.

ofedoren (approved):
Thanks, @qcjames53 ! We could probably generalize warning message/options instead of hardcoding 2 messages, but that can be dealt with in the future. As of now it seems and works quite nicely.

PR #396, opened 2025-05-14
Refs #38410 - Translate warning message

ofedoren:


ofedoren:
@qcjames53, sorry, I forgot to mention during review, we need to translate user facing messages, just FYI :)

qcjames53:
Aw man I thought about translations and then forgot to actually make it happen. Thank you for making this PR.

```
Warning: Field 'Informations sur le contenu/Content View Environments/Environnement de cycle de vie/Nom' is deprecated. Consider using 'Content Information/Content View Environments/LE Name' instead.
```
The 'Content Information/Content View Environments/LE Name' string should probably be translated as well. The best solution would be to pull the full field label from the new field. Alternatively we could just translate the replacement field string wherever used at the risk of the two strings being translated differently. I'll look into the former in the next hour or two.

ofedoren:
> The 'Content Information/Content View Environments/LE Name' string should probably be translated as well.

Fortunately, it is :) Unfortunately we don't have proper translations it seems:

Screenshot-1747309215783

PR #397, opened 2025-05-15
Fixes #38427 - Modify replaced_by to allow for selection of new field by object.

qcjames53:
Modified the `replaced_by` string on fields to a `replaced_by_path` relative path to another field. This allows for displaying the exact translation of the new path to stderr and can better handle label name changes in the future.

_From the docs:_
A field may be 'replaced by' another field using `:replaced_by_path => ["relative","path","to","new","field"]`. Use "!p" to indicate the parent field/list/collection and other strings to indicate the id of child field/list/collections. This will mark the field as deprecated and print a similar warning message to stderr whenever the field is displayed.

For example, to replace 'Foo/Bar/Old field' with 'Foo/Baz/New field', do the following:

```
field :old_fld, _("Old Field"), Fields::Field, :sets => ['ALL'], :replaced_by => ["!p","!p","baz","new_fld"]
```

Example output:

```
$ hammer foo info --fields "Foo/Bar/Old field"
Warning: Field 'Foo/Bar/Old field' is deprecated. Consider using 'Foo/Baz/New field' instead.
Foo:
  Bar:
    Old field: data
```

### DIFF
--- a/doc/creating_commands.md
+++ b/doc/creating_commands.md
@@ -755,7 +755,7 @@ def adapter
 end
 ```
 
-#### Deprecating fields
+#### Deprecating and replacing fields
 To deprecate a field, add `:deprecated => true` as an option for the field. This will print a warning message to stderr whenever the field is displayed. Consider removing this field from the default set so it is not displayed without a `--fields` param:
 
 ```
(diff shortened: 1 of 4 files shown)


qcjames53:
Closing PR as these changes are no longer considered the best approach.

PR #398, opened 2025-05-20
Fixes #38427 - Update field documentation to reflect replaced_by translation

qcjames53:
Update the 'Creating commands' documentation page to show the best practice of translating each segment of a field path individually when using 'replaced_by':

```
field ~~~~~ :replaced_by => [_('Bar'), _('New Field')].join('/')
```

### DIFF
--- a/doc/creating_commands.md
+++ b/doc/creating_commands.md
@@ -755,7 +755,7 @@ def adapter
 end
 ```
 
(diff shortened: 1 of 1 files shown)


ofedoren (approved):
Thanks, @qcjames53 !

